PR #206, opened 2019-01-07
Configured HttpClient to send User-Agent etc.

jmatsu:
## Issue
- None

## Overview (Required)
- Just for a purpose to collect stats later on API side :)

## Links
- None


takahirom:
Good job!

PR #212, opened 2019-01-07
Use binding inflate

STAR-ZERO:
## Issue
- None

## Overview (Required)
- Change to use inflate method of DataBinding class

## Links
- None

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
You are first contributer🏆

konifar:
I have a question! 🙋‍♂️ 
I thought `DataBindingUtil` is better because if we use `FragmentAboutBinding` to inflate layout, AndroidStudio detects `layyout/fragment_about.xml` as an unused resource. Is that correct?

takahirom:
Is this not affected? 👀  https://issuetracker.google.com/issues/116541301#comment17

STAR-ZERO:
Yes that is correct. When use `inflate` of binding class, layout xml is detected as unused. Actually I didn't notice it.
Using `DataBindingUtil` is better. However this return type is obvious.

According to document, both is OK.
https://developer.android.com/topic/libraries/data-binding/generated-binding

STAR-ZERO:
> Is this not affected? 👀 https://issuetracker.google.com/issues/116541301#comment17

No. It's different method signature. I don't use `DataBindingComponent`.

takahirom:
> No. It's different method signature. I don't use DataBindingComponent

Thanks! I misunderstood 👍

PR #214, opened 2019-01-07
fix directory path with to package name

satoshun:
## Issue
-  No issue

## Overview (Required)
- I think, directory path should match package name!

## Links
- Nothing

## Screenshot
Before | After
:--: | :--:
 |


konifar:
🦅 LGTM!

PR #223, opened 2019-01-07
Fix session time texts on session list not to be overlapped

### LINKED ISSUE #51: Fix item decoration for session time
## Overview (Required)
- current
decoration
 - Instead of disappearing when scrolling up, push out.
iosched2018

kafumi:
## Issue
- close #51

## Overview (Required)
- Fixed item decoration drawing session time texts on session list so that the texts are not overlapped

## Links
N/A

## Screenshot
Before | After
:--: | :--:
svid_20190108_001357_1 | svid_20190108_001100_1

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
@@ -23,8 +23,11 @@ class SessionsItemDecoration(
     private val textLeftSpace = resources.getDimensionPixelSize(
         R.dimen.session_bottom_sheet_left_time_text_left
     )
-    private val textMinTop = resources.getDimensionPixelSize(
-        R.dimen.session_bottom_sheet_left_time_text_top_min
+    private val textPaddingTop = resources.getDimensionPixelSize(
+        R.dimen.session_bottom_sheet_left_time_text_padding_top
+    )
+    private val textPaddingBottom = resources.getDimensionPixelSize(
+        R.dimen.session_bottom_sheet_left_time_text_padding_bottom
     )
 
     val paint = Paint().apply {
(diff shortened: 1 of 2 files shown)


feature/session/src/main/res/values/dimens.xml
```diff
@@ -3,7 +3,8 @@
     <dimen name="session_bottom_sheet_left_time_space">76dp</dimen>
     <dimen name="session_bottom_sheet_left_time_text_size">20dp</dimen>
     <dimen name="session_bottom_sheet_left_time_text_left">16dp</dimen>
-    <dimen name="session_bottom_sheet_left_time_text_top_min">22dp</dimen>
```
kafumi: I've removed `session_bottom_sheet_left_time_text_top_min` because it's not clear for what we should use it... If we should keep it for future use, please suggest it.

takahirom:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -45,37 +48,34 @@ class SessionsItemDecoration(
...
+            )
         }
     }
+
+    private fun getSessionTime(position: Int): String? {
+        if (position < 0 || position >= groupAdapter.itemCount) {
+            return null
+        }
+
+        val item = groupAdapter.getItem(position) as? SessionItem
```
takahirom: How about using elvis operator?♻️

groupAdapter.getItem(position) as? SessionItem ?: return null

kafumi: Sure! Actually, I was torn between current code and your proposal then selected current one... Will fix.

takahirom: Thanks! I will sleep 💤 Good night!

kafumi: Thank you for the swift review even midnight (or early morning)... Good night!

kafumi: Fixed at 4f54446d112ea4cdd233e9f9321583c9dad50e58.

PR #230, opened 2019-01-08
Introduce AppCoroutineDispatchers.

sys1yagi:
Hi! it's a great repo. I have gained a lot of idea from this repository! 🍰 

## Issue

none (Should I make one? 👀 )

## Overview (Required)

I think CoroutinePlugin may be a bit confusing.
For example, the following code is hard to imagine what happens.

https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/session/src/test/java/io/github/droidkaigi/confsched2019/session/ui/store/SessionContentsStoreTest.kt#L27
```
@Before fun setUp() {
  MockKAnnotations.init(this, relaxUnitFun = true)
  CoroutinePlugin.mainDispatcherHandler = { Dispatchers.Default } // what will happen?
}
```

So, I think that it would be better to declare an injectable dispatchers class.

### Introduce AppCoroutineDispatchers

AppCoroutineDispatchers is simple CoroutineDispatcher holder that using it, you can explicitly replace the Dispatcher during testing.

__pros__

You can explicitly specify Dispatcher.

__cons__

Some functions become ugly...

```
@MainThread fun <T> ReceiveChannel<T>.toLiveData(
    defaultValue: T? = null,
    appCoroutineDispatchers: AppCoroutineDispatchers // wow...
): LiveData<T>
```

## digest of changes

- Add AppCoroutineDispatchers
- Remove `CoroutinePlugin`, `io.github.droidkaigi.confsched2019.ext.android.Dispatchers`
- Replace the code that implicitly uses CoroutinePlugin to AppCoroutineDispatchers.

## Links

- original idea of [AppCoroutineDispatchers](https://github.com/chrisbanes/tivi/blob/master/base/src/main/java/app/tivi/util/AppCoroutineDispatchers.kt)

## Screenshot

none


sys1yagi:
What do you think about AppCoroutineDispatchers? @takahirom

takahirom:
Thank you for the suggestion!
I think that this is certainly an excellent way to do it. If the switching of Dispatcher is required frequently depending on the scope of Dagger, I think that this approach is reasonable.
However, since we are basically changing only in the test now, we believe that `toLiveData ()` etc is simpler and better for the current policy.

sys1yagi:
Thank you for your reviews!

>However, since we are basically changing only in the test now, we believe that toLiveData () etc is simpler and better for the current policy.

agree. 👍 

>We need to touch the root node of the DI graph for integration testings, i.e. we should know the root provider of the dispatcher holder in the test case

I lacked that point of view! :scream_cat:

sys1yagi:
> kotlinx.coroutine.corouting-testing

oh, yes! 

https://github.com/Kotlin/kotlinx.coroutines/blob/master/core/kotlinx-coroutines-test/README.md

`kotlinx-coroutines-test` can use `Dispatchers.setMain()`!
I'll try to use it 💪  and will make a new pull request at tonight.

sys1yagi:
close this pr and i'll make new pr later. 🌮

PR #238, opened 2019-01-08
Restore searched text

### LINKED ISSUE #176: Restore searched text when return from detail screen
## Overview (Required)
- When session -> search -> session detail -> back
- Searched text is cleard. We should restore it.
image

- We want to keep the inputted text
search

rei-m:
## Issue
- close #176

## Overview (Required)
- Restore searched text when return from detail screen. Add query string in searchStore and restored it when onCreateOptionsMenu.
- I'd like to confirm specification. I move query string to global store, so this case `session -> search -> session detail -> back -> back to session -> search again`  query string is restored, is it ok ?

## Links
- none

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
@@ -68,8 +68,8 @@ class SearchFragment : DaggerFragment() {
 
         sessionContentsStore.sessionContents.changed(viewLifecycleOwner) { contents ->
             searchActionCreator.search(
-                searchView?.query?.toString(),
-                sessionContentsStore.sessionContents.requireValue()
+                searchStore.searchResult.value?.query,
+                contents
             )
         }
         // TODO apply design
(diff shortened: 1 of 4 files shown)


takahirom:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
```diff
@@ -68,8 +68,8 @@ class SearchFragment : DaggerFragment() {
 
         sessionContentsStore.sessionContents.changed(viewLifecycleOwner) { contents ->
             searchActionCreator.search(
-                searchView?.query?.toString(),
-                sessionContentsStore.sessionContents.requireValue()
+                searchStore.searchResult.value?.query,
```
takahirom: Can you add query property for SearchStore?

```
    val query get() = searchResult.value?.query
```

rei-m: Done !! plz confrim commit https://github.com/DroidKaigi/conference-app-2019/pull/238/commits/99001bbf206b2c7be50895b55886ea0dfd6bc822

PR #244, opened 2019-01-08
replace session_label with search_label

tarumzu:
## Issue
- none

## Overview (Required)
-Timeline will be displayed when the search icon's long tap

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


panpanini:
could you add a link to the issue number as well? :pray:

takahirom:
Nice catch 🐱

takahirom:
I think the issue number is not needed 😉

tarumzu:
@panpanini sorry🙏. This issue was not made

Issue #136, opened 2019-01-05
Kotlin Multiplatform API implementation


panpanini:
## Overview (Required)
- In order to achieve #132, we should make the API implementation as a Kotlin Multiplatform library.

This will require separating pure Kotlin code from Java (eg: Dagger, etc), and bundling each into a jar, or a framework for Kotlin/Native

## Links
- https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html

kikuchy:
Just a question.
It there any plans to place iOS API implementation source codes?

Currently, Android API implementations are placed like below (ex. `data` directory).
It seems that we should make new modules like `api-impl-ios` separating from the Android implementation module.

```
data
├── api  (interfaces only)
│   └── src
│       ├── androidTest
│       ├── commonMain
│       ├── main
│       └── test
├── api-impl  (Android implementation only)
│   └── src
│       ├── main
│       └── test
```

JetBrains's example contains Android and iOS sources into one module (to provide `actual` implementations).  
https://github.com/JetBrains/kotlin-mpp-example/tree/master/greeting/src  
And we will be able to place iOS implementations like `data/api-impl/iosMain`

Which way repository owners are assuming?

takahirom:
If we can, I want to do `data/api-impl/iosMain`.

kikuchy:
I see. Thank you for replying!

takahirom:
done. 👍

PR #246, opened 2019-01-08
Make `api-impl` to Kotlin Multiplatform module

kikuchy:
## Issue
- A step of #136

## Overview (Required)
- Make `:data:api-impl` to Kotlin Multiplatform module.
- Keep injectable API clients constructers.
- No iOS implementation yet.
    - It's difficult for me to get the value of `API_ENDPOINT` in build.gradle from iOS implementation.
    - Please give me some ideas 😢

## Links
- https://kotlinlang.org/docs/tutorials/native/mpp-ios-android.html
- https://ktor.io/clients/http-client/features/user-agent.html


takahirom:
👀

takahirom:
Probably I feel that passing ENDPOINT is good 👍 (But If we can find good solution for it, I want to replace.)

jmatsu:
I know BuildConfig is android-specific... 🙇  
I'm not sure how to get such build-time conditional values on iOS/web/Native of multi-platform projects but it would be okay to declare a common class which has api endpoint values if possible.

I'm a more newbie of multiplatform than other newbies... so this is just an idea.
If we can use macro for iOS project and write a binding to Kotlin's actual implementation, then the code would be like below.

```
// common
internal expect fun apiEndpoint(): String
```

```
// android
internal actual fun apiEndpoint(): String {
    return BuildConfig.API_ENDPOINT
}

// ios
internal actual fun apiEndpoint(): String {
    return <endpoint from obj c>
}
```

takahirom:
Any way we can merge this!
If you add iOS code, You can consider this 👍 
https://github.com/DroidKaigi/conference-app-2019/pull/246#issuecomment-452247536

takahirom:
We are waiting for your next contribution 🥇

kikuchy:
Thanks a lot!

I will take this https://github.com/DroidKaigi/conference-app-2019/pull/246#issuecomment-452247536 way if I can implement...!

PR #247, opened 2019-01-08
Fix bug of using different store

takahirom:
## Overview (Required)
- Some time filter is not applied 😭 

## Screenshot
Before | After
:--: | :--:
before | after


takahirom:
let me merge 🔧

PR #267, opened 2019-01-08
remove unnecessary null-check

### LINKED ISSUE #266: unnecessary safe call on a non-null - 不必要なNullチェックしている
## Overview (Required)
- unnecessary null check

https://github.com/DroidKaigi/conference-app-2019/blob/ad2ec2e9a6ed39159204d7f649d8e6aef4ea7b62/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/item/ServiceSessionItem.kt#L23

fumiya-kume:
## Issue
- close #266

## Overview (Required)
- remove unnecessary null-check


takahirom:
Nice catch 🐱

PR #273, opened 2019-01-08
Add "about screen" to content RecyclerView item Preview

fumiya-kume:
## Issue
- none

## Overview (Required)
- Add "about screen" to content RecyclerView item Preview
- I think best is preview header and item, but now preview item only. (HELP plz... :D

## Screenshot
// TODO upload image after send PR
Before | After
:--: | :--:
 |


takahirom:
Thank you for cool preview 🆒

PR #293, opened 2019-01-09
Enable offline build on the CI

jmatsu:
## Issue
- #305 

## Overview (Required)
- I know people love faster builds.
- `dependencies` does not download all dependencies, and also nothing will be downloaded as for this project.
- `androidDependencies` does not contain kapt and custom configurations
- So we need a task to download kapt and custom configurations. After that, let's combine it with `androidDependencies`.
- A html reporter was downloaded every time but it was not used on CI

## Links
-

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,7 @@ jobs:
           keys: *all_keys_of_gradle_cache
       - run:
           name: Download Dependencies
-          command: retry_command ./gradlew androidDependenciesExtracs getDependencies
+          command: retry_command ./gradlew androidDependenciesExtra getDependencies
       - save_cache: &save_gradle_cache
           paths:
             - ~/.android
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -8,13 +8,11 @@ cache_keys:
     keys: &all_keys_of_gradle_cache
       - *primary_key_of_gradle_cache
       - gradle-cache-{{ checksum "~/CACHE_VERSION_OF_PROJECT_DEPS" }}-
-      - gradle-cache-
```
jmatsu: It should be aligned with the major cache version.

jmatsu:
yay!!!

.circleci/scripts/retry_command
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
```
takahirom: 🆒

build.gradle
```diff
@@ -68,6 +76,14 @@ allprojects {
                     main = "com.github.shyiko.ktlint.Main"
                     args "--android", '-F', 'src/main/**/*.kt', 'src/debug/**/*.kt', 'src/release/**/*.kt'
                 }
+
+                task androidDependenciesExtra(dependsOn: 'androidDependencies') {
```
takahirom: This task is very useful 🎉

PR #296, opened 2019-01-09
Fix to hide BottomAppBar when session detail can not scroll

### LINKED ISSUE #153: BottomAppBar does not disappear on session details that can not be scrolled
## Overview (Required)
- BottomAppBar does not disappear on session details that can not be scrolled

image

androhi:
## Issue
- close #153 

## Overview (Required)
- Added SessionBottomAppBarBehavior class
- If screen be able not to scrolled, it makes to slide the bottom app bar

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionBottomAppBarBehavior.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionBottomAppBarBehavior.kt
@@ -11,11 +11,15 @@ import com.google.android.material.bottomappbar.BottomAppBar
 class SessionBottomAppBarBehavior(
     context: Context? = null,
     attrs: AttributeSet? = null
-    ) : BottomAppBar.Behavior(context, attrs) {
+) : BottomAppBar.Behavior(context, attrs) {
 
     private var state: Int = STATE_SCROLLED_UP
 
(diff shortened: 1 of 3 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionBottomAppBarBehavior.kt
```diff
@@ -0,0 +1,57 @@
...
+) : BottomAppBar.Behavior(context, attrs) {
+
+    private var state: Int = STATE_SCROLLED_UP
+
+    override fun onInterceptTouchEvent(
+        parent: CoordinatorLayout,
+        child: BottomAppBar,
+        ev: MotionEvent
+    ): Boolean {
+        if (child.isShown.not() || ev.action != MotionEvent.ACTION_DOWN) {
```
takahirom: We can discuss in the issue. But Currently, We are using `!` instead of `.not()` 🙏

androhi: I see! Thanks for your review.

androhi: fixed! 5c9b771

takahirom:
I confirm that! It works perfectly! Thanks 👍

PR #299, opened 2019-01-09
Reduce the bottom touch area of the session item

### LINKED ISSUE #181: Expand top touch area for session item top and bottom
## Overview (Required)
- I want to fix only the tap area without changing the appearance.
- Expand top touch area for session item top and bottom. We can reduce the bottom touch area.
image
- Please be aware that time must come at the same position as the title.
image

yurihondo:
## Issue
 #181 

## Overview (Required)
The title of the issue (#181 ) is "Expand top touch area...", but since there is description "We can reduce the bottom touch area" in the comment, I implemented it accordingly.(Please check **Screenshot**.)

## Links
none

## Screenshot
Before | After
:--: | :--:
 |


yurihondo:
Could you please point it out if this change is different from the intent of the issue?

takahirom:
Thanks! It's better.
I want to merge this without remove issue.

My ideal is like this form.
image

yurihondo:
@takahirom 
Thank you for detailed explanation. I get the point.

>I want to merge this without remove issue.
OK. Please merge this PR.
I will create an another PR for the layout change which you have taught.

takahirom:
Thanks 🥈

PR #301, opened 2019-01-09
Close drawer when pressed back key while drawer opened.

sobachanko:
## Issue
- none

## Overview (Required)
- According to title.

## Screenshot
Before | After
:--: | :--:
close_drawer_before | close_drawer_after


takahirom:
Thanks 👍I forgot this.

PR #302, opened 2019-01-09
Fixed the width of divider in session detail screen

NUmeroAndDev:
## Overview (Required)
- Fixed the width of divider in session detail screen to be the same as gallery.
In gallery, the right margin of the session detail screen is 20dp, but according to the implementation's 16 dp. 🤔 

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Thanks! We need this 👍

PR #303, opened 2019-01-09
 Suppress Fragment replacement

### LINKED ISSUE #33: Keep scroll position when back from session detail
# Overview
When back button press in session detail, scroll position will become top.
Because of BottomSheetDaySessionsFragment and BottomSheetFavoriteSessionsFragment instances are recreated.
If you know how to fix this, Please help me 🆘

Akihiroooo:
## Issue
- close #33 

## Overview (Required)
- Suppress BottomSheetFragment replacement.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
@@ -70,7 +70,9 @@ class SessionPageFragment : DaggerFragment() {
 
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
-        setupSessionsFragment(savedInstanceState)
+        if (savedInstanceState == null) {
+            setupSessionsFragment()
+        }
     }
 
     override fun onCreateView(
@@ -125,11 +127,7 @@ class SessionPageFragment : DaggerFragment() {
         }
     }
 
-    private fun setupSessionsFragment(savedInstanceState: Bundle?) {
-        if (savedInstanceState != null) {
-            return
-        }
-
+    private fun setupSessionsFragment() {
         val tab = SessionPage.pages[args.tabIndex]
         val fragment: Fragment = when (tab) {
             is SessionPage.Day -> {
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
@@ -136,10 +136,11 @@ class SessionPageFragment : DaggerFragment() {
             }
 
             // suppress fragment replacement
-            if (childFragmentManager.findFragmentByTag(fragment::javaClass.name) == null) {
+            val title = SessionPage.pages[args.tabIndex].title
+            if (childFragmentManager.findFragmentByTag(title) == null) {
                 childFragmentManager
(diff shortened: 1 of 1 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
```diff
@@ -135,11 +135,14 @@ class SessionPageFragment : DaggerFragment() {
                 }
             }
 
-            childFragmentManager
-                .beginTransaction()
-                .replace(R.id.sessions_sheet, fragment)
-                .disallowAddToBackStack()
-                .commit()
+            // suppress fragment replacement
+            if (childFragmentManager.findFragmentByTag(fragment::javaClass.name) == null) {
```
takahirom: I am worried that Day Fragment's class is the same on Day 1 and Day 2, so when the Fragment is reused by ViewPager, information on another tab will be displayed.
So how about using the SessionPage title as a tag? 👀

Akihiroooo: so cool!! 👍 
I'll do it!

takahirom: And probably we can put Fragment creation in if statement.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
```diff
@@ -121,8 +121,10 @@ class SessionPageFragment : DaggerFragment() {
     }
 
     private fun setupBottomSheet(savedInstanceState: Bundle?) {
-        if (savedInstanceState == null) {
-            val fragment: Fragment = when (val tab = SessionPage.pages[args.tabIndex]) {
+        // suppress fragment replacement
+        val tab = SessionPage.pages[args.tabIndex]
+        if (savedInstanceState == null && childFragmentManager.findFragmentByTag(tab.title) == null) {
```
Akihiroooo: > And probably we can put Fragment creation in if statement

I may not understand this sentence's intention correctly. If so, I'm sorry.

takahirom:
I'm thinking about this 👀 
https://twitter.com/swiz_ard/status/1082943330852397056

takahirom:
Can you try it? 🙏

Akihiroooo:
Since the method `setupBottomSheet()` relies on data binding, it is assumed that an uninitialized exception will be raised. Therefore, method `setupSessionsFragment()` has been added.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
```diff
@@ -68,6 +68,11 @@ class SessionPageFragment : DaggerFragment() {
     private val bottomSheetBehavior: BottomSheetBehavior<*>
         get() = BottomSheetBehavior.from(binding.sessionsSheet)
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setupSessionsFragment(savedInstanceState)
```
jmatsu: `savedInstanceState == null` check should be here rather than inside of * setupSessionsFragment* 👀

takahirom:
Thank you finding solution! 🏆

PR #304, opened 2019-01-09
Fix hide keyboard when search result

### LINKED ISSUE #291: When open session/speaker detail from search result, Keyboard is not hidden
## Overview (Required)
- title says everything
image

kaya5777:
## Issue
- close #291

## Overview (Required)
- hide keyboard when search result

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
@@ -135,7 +135,6 @@ class SearchFragment : DaggerFragment() {
     override fun onPause() {
         super.onPause()
 
-        android.util.Log.e("hoge", "hoge")
         val imm = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
             context?.systemService<InputMethodManager>()
         } else {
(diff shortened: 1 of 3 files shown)


takahirom:
Thanks 👍 
If I introduce this, I should implement this for all screen. 😭 
Can you do it in only SearchFragment?

kaya5777:
Sorry, I understand.
I fix only SearchFragment to solve this problem.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
```diff
@@ -127,6 +131,17 @@ class SearchFragment : DaggerFragment() {
             searchView.setOnCloseListener { false }
         }
     }
+
+    override fun onPause() {
+        super.onPause()
+
+        val imm = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
```
jmatsu: I'm very sorry, I've forgotten this api requires Android M... so `ContextCompat.getSystemService(requireContext(), InputMethodService::class.java)` looks better. 🙇

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SearchFragment.kt
```diff
@@ -127,6 +131,17 @@ class SearchFragment : DaggerFragment() {
...
+            context?.systemService<InputMethodManager>()
+        } else {
+            context?.getSystemService(Context.INPUT_METHOD_SERVICE) as? InputMethodManager
+        }
+        imm?.hideSoftInputFromWindow(binding.root.windowToken, 0)
```
jmatsu: please use the currently focused view instead.

PR #321, opened 2019-01-09
Delete language resources that are not Japanese and English by resConfigs

tatsuhama:
## Issue
- None

## Overview (Required)
- Delete language resources that are not Japanese and English by resConfigs
- Support libraries includes string resources other than Japanese and English
- Effect : Reduce apk size

## Links
- https://developer.android.com/studio/build/shrink-code?#unused-alt-resources
- https://qiita.com/tatsuhama/items/814471b79c5d572f77e9
- https://github.com/DroidKaigi/conference-app-2018/pull/129

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/frontend/android/build.gradle
+++ b/frontend/android/build.gradle
@@ -25,7 +25,7 @@ android {
         versionName "1.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         multiDexEnabled true
-        resConfigs "en", "ja"
+        resConfigs "ja"
         // For Android Studio Gradle sync
         applicationIdSuffix ".debug"
     }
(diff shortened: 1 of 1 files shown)


takahirom:
Sorry, We are checking CI. 🙇

frontend/android/build.gradle
```diff
@@ -25,6 +25,7 @@ android {
         versionName "1.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         multiDexEnabled true
+        resConfigs "en", "ja"
```
jmatsu: `en` is needless because `en` resource does not exist in this app.
English materials are in the default resources so `resConfigs "ja"` is enough as for this app. 👍 

To check the difference, it would be nice if you can compare between each result of `aapt dump badging $apk_path`.

tatsuhama:
@jmatsu 
Thank you for comment.
I checked and fixed.
Please check again 🙏

jmatsu:
Let me re-run the CI job just in case because it has failed.

PR #325, opened 2019-01-09
Share session

### LINKED ISSUE #121: Session share
## Overview (Required)
- Currently, the share feature is not implemented.
image
- Blocked by web (URL)

h3birth:
## Issue
- close #121 

## Overview (Required)
- add share feature on session detail page
- The shared url text changes according to the language setting of Android OS. 
   - For example when Japanese "https://droidkaigi.jp/2019/timetable/12345/" , other "https://droidkaigi.jp/2019/en/timetable/12345/"

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -68,12 +68,13 @@ class SessionDetailFragment : DaggerFragment() {
...
-                    Toast.makeText(
-                        requireContext(),
-                        "not implemented yet",
-                        Toast.LENGTH_SHORT
-                    ).show()
+                R.id.session_share -> {
+                    val session = binding.session ?: return@setOnMenuItemClickListener false
+                    activityActionCreator.shareUrl(getString(
+                        R.string.session_detail_share_url,
+                        session.id.toInt()
```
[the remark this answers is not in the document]
takahirom: I don't have this solution
Can you add this?

```
        @Suppress("StringFormatMatches") // FIXME
```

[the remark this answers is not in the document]
h3birth: Yes, added ` @Suppress("StringFormatMatches") // FIXME`

feature/system/src/main/java/io/github/droidkaigi/confsched2019/system/actioncreator/ActivityActionCreator.kt
```diff
@@ -34,4 +35,11 @@ class ActivityActionCreator @Inject constructor(val activity: FragmentActivity)
         const val LATITUDE_LOCATION = "35.696065"
         const val LONGITUDE_LOCATION = "139.690426"
     }
+
+    fun shareUrl(url: String) {
```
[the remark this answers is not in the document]
h3birth: I'm sorry, I fixed it.

h3birth:
@takahirom Thank you review. 
I fixed that. Please review again. 🙇

feature/session/src/main/res/values/strings.xml
```diff
@@ -17,6 +17,7 @@
     <string name="session_detail_resource">Resources</string>
     <string name="session_detail_resource_video">Video</string>
     <string name="session_detail_resource_slide">Slides</string>
+    <string name="session_detail_share_url">https://droidkaigi.jp/2019/en/timetable/%d/</string>
```
takahirom: It's a very small thing 🙇 
Can you delete last `/` ?

jmatsu: 👀 
We are sorry, examples we provided were only integer but some of session ids are String. 
So this format argument should be `%s` instead.

h3birth: Thank you. I fixed delete last / and change format argument.

h3birth:
@takahirom  Thank you for review many times.
I fixed that. Please review again. 🙇

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -68,12 +68,14 @@ class SessionDetailFragment : DaggerFragment() {
...
             when (item.itemId) {
-                R.id.session_share ->
-                    Toast.makeText(
-                        requireContext(),
-                        "not implemented yet",
-                        Toast.LENGTH_SHORT
-                    ).show()
+                R.id.session_share -> {
+                    val session = binding.session ?: return@setOnMenuItemClickListener false
+                    @Suppress("StringFormatMatches") // FIXME
```
panpanini: > // FIXME

🤔 😹

feature/session/src/main/res/values-ja/strings.xml
```diff
@@ -8,6 +8,7 @@
     <string name="session_detail_resource">資料</string>
     <string name="session_detail_resource_video">動画</string>
     <string name="session_detail_resource_slide">スライド</string>
+    <string name="session_detail_share_url">https://droidkaigi.jp/2019/timetable/%s</string>
```
panpanini: I don't think we need a translation for this 👍

feature/session/src/main/res/values/strings.xml
```diff
@@ -17,6 +17,7 @@
     <string name="session_detail_resource">Resources</string>
     <string name="session_detail_resource_video">Video</string>
     <string name="session_detail_resource_slide">Slides</string>
+    <string name="session_detail_share_url">https://droidkaigi.jp/2019/en/timetable/%s</string>
```
panpanini: if you add the `translatable="false"` parameter to this, then the warning will go away 😉 
``` xml
    <string name="session_detail_share_url" translatable="false">https://droidkaigi.jp/2019/en/timetable/%s</string>
```

feature/session/src/main/res/values-ja/strings.xml
panpanini: ah, actually maybe we do 👀 because this should transition to Japanese session

h3birth: Oh, i see. Thank you so much. :smile:

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
h3birth: I'm sorry, i didn't understand. I fixed it 🙇

feature/session/src/main/res/values-ja/strings.xml
panpanini: Sorry @h3birth  :sob: I made a mistake here, could you add this URL back? because the URLs are different between English and Japanese, we should also have this URL in `values-ja` as well 🙇

h3birth:
@panpanini Thank you for review.
I fixed that. Please review again. 🙇

panpanini:
Thanks! it looks like CI failed, so I'll re-run it 🙇

PR #330, opened 2019-01-09
Fix SessionsItemDecoration not to draw unexpected time text

### LINKED ISSUE #324: Unexpected time text appears on session list when toggling fav status
## Overview (Required)

- On session list, when changing session favorite status, unexpected time text can appear for a moment
- This issue can be always reproduced if
  - the time frame has multiple session AND
  - change favorite status of the session listed last in the time frame

svid_20190109_223459_1

## Steps to reproduce

1. Launch app
2. On "Day1" session list, scroll to the "Unit test for ViewModel and LiveData" session
3. Click the favorite button of the session

## Expected behavior

- After step 3, there is no update about session time display

## Actual behavior

- After step 3, time text "11:20" appears on the left of the session even though another "11:20" text exists above

kafumi:
## Issue
- close #324

### Cause of the issue

- Time texts on session list are drawn by `SessionsItemDecoration`
- `SessionsItemDecoration` processes `RecyclerView`'s children ordered by index

```kotlin
        var lastTime: String? = null
        for (i in 0 until parent.childCount) {
            val view = parent.getChildAt(i)
            val position = parent.getChildAdapterPosition(view)
            if (position == -1 || position >= groupAdapter.itemCount) return

            val time = getSessionTime(position) ?: continue

            if (lastTime == time) continue
            lastTime = time
```

- It skips drawing text if the session time of **previous view** is same
- However, view index order and item position order can be different in RecyclerView
  - i.e. Drawing text should be skipped if the session time of **previous item** (not of previous view) is same
- Although they are luckily matched in most cases, updating item can cause the "unmatched" situation because it adds new View without recycling
  - In that case, `if (lastTime == time)` in above code doesn't work expectedly

### Solution

- Sort `RecyclerView`'s children by item position before processing
  - After this sort, we can assume `lastTime` has the time of previous position items
- Note: We can NOT remove `lastTime` usage by, for example, following code
  - It's because we need to make sure previous position item is in the visible range of `RecyclerView` to skip drawing text
    - If previous item is not visible, the time text has not been drawn. Text drawing shouldn't skip
  - Although following code performs comparison with previous item as expected, it doesn't assure previous item is in `RecyclerView`'s visible range

```kotlin
            val time = getSessionTime(position) ?: continue

            val prevItemTime = getSessionTime(position - 1)
            if (prevItemTime == time) continue
```

## Links
N/A

## Screenshot
Before | After
:--: | :--:
svid_20190109_223459_1 | svid_20190109_233031_1

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
@@ -5,7 +5,10 @@ import android.content.res.Resources
 import android.graphics.Canvas
 import android.graphics.Color
 import android.graphics.Paint
+import android.util.SparseArray
+import android.view.View
 import androidx.core.content.res.ResourcesCompat
+import androidx.core.util.forEach
 import androidx.recyclerview.widget.RecyclerView
 import com.xwray.groupie.GroupAdapter
 import io.github.droidkaigi.confsched2019.session.R
@@ -42,16 +45,24 @@ class SessionsItemDecoration(
         }
     }
 
+    // Keep SparseArray instance on property to avoid object creation in every onDrawOver()
+    private val positionToViews = SparseArray<View>()
+
     override fun onDrawOver(c: Canvas, parent: RecyclerView, state: RecyclerView.State) {
-        var lastTime: String? = null
+        // Sort child views by adapter position
         for (i in 0 until parent.childCount) {
             val view = parent.getChildAt(i)
             val position = parent.getChildAdapterPosition(view)
-            if (position == -1 || position >= groupAdapter.itemCount) return
+            if (position != -1 && position < groupAdapter.itemCount) {
+                positionToViews.put(position, view)
+            }
+        }
 
-            val time = getSessionTime(position) ?: continue
+        var lastTime: String? = null
+        positionToViews.forEach { position, view ->
+            val time = getSessionTime(position) ?: return@forEach
 
-            if (lastTime == time) continue
+            if (lastTime == time) return@forEach
             lastTime = time
 
             val nextTime = getSessionTime(position + 1)
@@ -68,6 +79,8 @@ class SessionsItemDecoration(
                 paint
             )
         }
+
+        positionToViews.clear()
     }
 
     private fun getSessionTime(position: Int): String? {
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
@@ -32,6 +32,8 @@ class SessionsItemDecoration(
     private val textPaddingBottom = resources.getDimensionPixelSize(
         R.dimen.session_bottom_sheet_left_time_text_padding_bottom
     )
+    // Keep SparseArray instance on property to avoid object creation in every onDrawOver()
+    private val adapterPositionToViews = SparseArray<View>()
 
     val paint = Paint().apply {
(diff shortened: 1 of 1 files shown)


takahirom:
Sorry, I’m gonna go to bed now. 💤

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -42,16 +46,24 @@ class SessionsItemDecoration(
         }
     }
 
+    // Keep SparseArray instance on property to avoid object creation in every onDrawOver()
+    private val positionToViews = SparseArray<View>()
```
jmatsu: `adapterPositionToViews` is preferable rather than `positionToViews` in my opinion. What do you think about it?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -42,16 +45,24 @@ class SessionsItemDecoration(
         }
     }
 
+    // Keep SparseArray instance on property to avoid object creation in every onDrawOver()
```
jmatsu: Make sense 👍

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
jmatsu: It seems other `private val` fields are above `public val`. Could you please sort fields? 👀

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -5,7 +5,11 @@ import android.content.res.Resources
 import android.graphics.Canvas
 import android.graphics.Color
 import android.graphics.Paint
+import android.util.SparseArray
+import android.util.SparseIntArray
```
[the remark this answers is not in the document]
kafumi: Fixed and squashed to d554a67c71f6b0ab43090e3b4f82e2ae2ab050a4.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -42,16 +46,24 @@ class SessionsItemDecoration(
...
+    private val positionToViews = SparseArray<View>()
+
     override fun onDrawOver(c: Canvas, parent: RecyclerView, state: RecyclerView.State) {
-        var lastTime: String? = null
+        // Sort child views by adapter position
         for (i in 0 until parent.childCount) {
             val view = parent.getChildAt(i)
             val position = parent.getChildAdapterPosition(view)
-            if (position == -1 || position >= groupAdapter.itemCount) return
+            if (position != -1 && position < groupAdapter.itemCount) {
```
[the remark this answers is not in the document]
kafumi: Fixed as suggested at 49e58360076432e53ecdbe3c73ea12f9a77e780c.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
kafumi: Included both suggestion at 49e58360076432e53ecdbe3c73ea12f9a77e780c.

kafumi:
@jmatsu Thank you for the review. I've addressed all suggestions. Please take a look when you have time :pray:

PR #331, opened 2019-01-09
Use Timber mpp

### LINKED ISSUE #274: use Timber Mpp?
## Overview (Required)

timber supports multi platform from 5.0.0-SNAPSHOT. How about using it?

pros
- Timber Android ecosystem is so large.
- Timber has a flexible API.

cons
- snapshot... not stable.
- iOS sample implements not exists in Timber project. So we must be implemented code for iOS Timber log.

## Links
- Timber: https://github.com/JakeWharton/timber
- SdkSearch: https://github.com/JakeWharton/SdkSearch

satoshun:
## Issue
- close #274 

## Overview (Required)
- Timber support MPP from 5.0.0 snapshot, i try it!

## Links
- https://github.com/JakeWharton/timber

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/ext/log/build.gradle
+++ b/ext/log/build.gradle
@@ -12,10 +12,12 @@ kotlin {
     sourceSets {
         commonMain.dependencies {
             api Dep.Kotlin.stdlibCommon
+            api Dep.Timber.common
         }
         androidMain {
             dependencies {
                 api Dep.Kotlin.stdlibJvm
+                api Dep.Timber.jdk
             }
         }
         commonTest.dependencies {
--- a/ext/log/src/commonMain/kotlin/io/github/droidkaigi/confsched2019/timber/Timber.kt
+++ b/ext/log/src/commonMain/kotlin/io/github/droidkaigi/confsched2019/timber/Timber.kt
@@ -1,0 +1,29 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
+package io.github.droidkaigi.confsched2019.timber
+
+import timber.log.Timber
+
+inline fun Timber.assert(throwable: Throwable) {
+    Timber.log(ASSERT, null, throwable, null)
+}
+
+inline fun Timber.error(throwable: Throwable) {
+    Timber.log(ERROR, null, throwable, null)
+}
+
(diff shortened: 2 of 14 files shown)


ext/log/src/commonMain/kotlin/io/github/droidkaigi/confsched2019/timber/Timber.kt
```diff
@@ -0,0 +1,29 @@
+@file:Suppress("NOTHING_TO_INLINE")
+
+package io.github.droidkaigi.confsched2019.timber
+
+import timber.log.Timber
+
+inline fun Timber.assert(throwable: Throwable) {
```
satoshun: Timber doesn't have helper method without message String. i think, it's helper method useful.

feature/session/build.gradle
```diff
@@ -49,6 +48,8 @@ dependencies {
     implementation Dep.Picasso.picasso
     implementation Dep.Picasso.picassoTransformation
 
+    implementation Dep.Timber.android
```
satoshun: compile error is occur if it's removed. but i add dependency of Timber in ":ext:log" module. kotlin-native doesn't support Gradle dependency api?

ext/log/build.gradle
```diff
@@ -12,10 +12,12 @@ kotlin {
     sourceSets {
         commonMain.dependencies {
             api Dep.Kotlin.stdlibCommon
+            api Dep.Timber.common
         }
         androidMain {
             dependencies {
                 api Dep.Kotlin.stdlibJvm
+                api Dep.Timber.jdk
```
satoshun: i add Timber dependency in it. but client module needs explicitly Timber dependency.

ext/log/src/commonMain/kotlin/io/github/droidkaigi/confsched2019/timber/Timber.kt
takahirom: It's useful!!

takahirom:
LGTM 🎄

PR #340, opened 2019-01-10
Applied a bit of modification to README

jmatsu:
## Issue
- None

## Overview (Required)
- `Google Play Protect will show a warning` -> `Google Play Protect will show a warning dialog`
- Added a line return

## Links
-


mhidaka:
Thank you @jmatsu

PR #343, opened 2019-01-10
Feature/189 show service session details

### LINKED ISSUE #189: Show codelabs details when touch codelabs items like normal sessions
## Overview (Required)
- When you touch codelabs items, nothing happens!

device-2019-01-07-001021

pocket7878:
## Issue
- close #189 

## Overview (Required)
- Separate ServiceSessionDetailFragment from SpeechSessionDetailFragment


model/src/commonMain/kotlin/SessionType.kt
```diff
@@ -14,4 +14,10 @@ enum class SessionType(val id: String) {
     companion object {
         fun of(id: String?) = values().find { it.id == id } ?: Unknown
     }
+
+    val supportDetail: Boolean
```
takahirom: Can you add the main constructor property?
 `SessionType(val id: String, val supportDetail: Boolean)`
Because when add SessionType, we forget that property 💭

takahirom:
Sorry, Currently master CI is failing. We are investigating. Please wait 🙏

takahirom:
The layout of the service session has already become a bit older.
I believe that one of the screens will not be updated like that in the future. Therefore, I think that if possible we do not want to separate process and layout.
I think you can implement it by using Binding 3 fields(Session, SpeechSession, ServiceSession).

pocket7878:
Thank you for your review! @takahirom 
I'll recreate my commit from latest master!

takahirom:
SessionDummyDatas.kt: (20, 13): Type mismatch: inferred type is Room but String was expected
Please fix SessionDummyDatas 🙏

takahirom:
Can I fetch your commit and merge it? 
I want to merge this. But conflicting 😭

PR #342, opened 2019-01-10
Implemented session tab animation

NUmeroAndDev:
## Issue
- close #328 

## Overview (Required)
- Implemented session tab animation.
But, RippleEffect is not the same as before. 
It was difficult for me. 😢 

## Links
- https://storage.googleapis.com/spec-host-backup/mio-design%2Fassets%2F1ZL4CiG4FN1MxyTuSqZqU_xX_V5pDakCs%2Fnavtransitions-peer.mp4

## Screenshot
Before | After
:--: | :--:
 |

PR #347, opened 2019-01-10
Fix Tab size when Landscape

NUmeroAndDev:
## Issue
- NONE

## Overview (Required)
- Fix Tab size when Landscape.
I do not know the design of Landscape, is it too big? 🤔 

## Links
- #342 

## Screenshot
Before | After
:--: | :--:
 |


mhidaka:
Thanks @NUmeroAndDev

PR #353, opened 2019-01-10
Fix time format to 24-hour clock

trashfeed:
## Issue
- None.

## Overview (Required)
- switch views between a 24-hour clock at session detail page.

## Links
- None.

## Screenshot
Before | After
:--: | :--:
 | 
- session list page is 24-hour clock, there  is no problem.


e10dokup:
LGTM 🎉

PR #362, opened 2019-01-10
Set maven urls based on the current regions

### LINKED ISSUE #361: Use Google Mirror Maven Central based on the region
## Overview (Required)

- Google Mirror Maven Central could reduce the sync time (see https://github.com/DroidKaigi/conference-app-2019/issues/168) though, CI region might be unknown so we cannot use suitable ones. 
- Now we might be able to use ipinfo.io and country.to for this stuff. https://ipinfo.io/country can return the country of the caller and http://country.io/continent.json is an index of regions by countries.
- In short, we may be able to select a suitable mirror somehow.

## Links
- Related to https://github.com/DroidKaigi/conference-app-2019/issues/168

jmatsu:
## Issue
- close #361

## Overview (Required)
- This is just an idea based on the caller IP based region estimate
- I'm wondering where should be the default region... 🤔 

## Links
-


scripts/bash.source
```diff
@@ -1,3 +1,9 @@
 export PATH="${REPOSITORY_ROOT}/scripts:$PATH"
 export PATH="${REPOSITORY_ROOT}/.circleci/scripts:$PATH"
-export MERGED_JUNIT_RESULT_FILE="$REPOSITORY_ROOT/merged_junit.xml"
\ No newline at end of file
+export MERGED_JUNIT_RESULT_FILE="$REPOSITORY_ROOT/merged_junit.xml"
+
+export CURRENT_COUNTRY=$(curl 'https://ipinfo.io/country')
```
takahirom: I don't know this API! Can we use this free? Not limit counts? 👀

jmatsu: free, and the call limit is 1000 times per day from the same IP. So it can be assumed as no limit seemingly.

.circleci/continent.json
```diff
@@ -0,0 +1 @@
... "CM": "AF", "CL": "SA", "CC": "AS", "CA": "NA", "CG": "AF", "CF": "AF", "CD": "AF", "CZ": "EU", "CY": "EU", "CX": "AS", "CR": "NA", "CW": "NA", "CV": "AF", "CU": "NA", "SZ": "AF", "SY": "AS", "SX": "NA", "KG": "AS", "KE": "AF", "SS": "AF", "SR": "SA", "KI": "OC", "KH": "AS", "KN": "NA", "KM": "AF", "ST": "AF", "SK": "EU", "KR": "AS", "SI": "EU", "KP": "AS", "KW": "AS", "SN": "AF", "SM": "EU", "SL": "AF", "SC": "AF", "KZ": "AS", "KY": "NA", "SG": "AS", "SE": "EU", "SD": "AF", "DO": "NA", "DM": "NA", "DJ": "AF", "DK": "EU", "VG": "NA", "DE": "EU", "YE": "AS", "DZ": "AF", "US": "NA", "UY": "SA", "YT": "AF", "UM": "OC", "LB": "AS", "LC": "NA", "LA": "AS", "TV": "OC", "TW": "AS", "TT": "NA", "TR": "AS", "LK": "AS", "LI": "EU", "LV": "EU", "TO": "OC", "LT": "EU", "LU": "EU", "LR": "AF", "LS": "AF", "TH": "AS", "TF": "AN", "TG": "AF", "TD": "AF", "TC": "NA", "LY": "AF", "VA": "EU", "VC": "NA", "AE": "AS", "AD": "EU", "AG": "NA", "AF": "AS", "AI": "NA", "VI": "NA", "IS": "EU", "IR": "AS", "AM": "AS", "AL": "EU", "AO": "AF", "AQ": "AN", "AS": "OC", "AR": "SA", "AU": "OC", "AT": "EU", "AW": "NA", "IN": "AS", "AX": "EU", "AZ": "AS", "IE": "EU", "ID": "AS", "UA": "EU", "QA": "AS", "MZ": "AF"}
```
takahirom: Why is this in .circleci?

build.gradle
```diff
@@ -32,7 +32,24 @@ buildscript {
 allprojects {
     repositories {
         google()
-        if (!isCi) { maven { url "https://maven-central-asia.storage-download.googleapis.com/repos/central/data/" } }
+
+        // to know your region, please have a look at scripts/bash.source
```
takahirom: 👍

jmatsu:
I'd love to do benchmark though, the current results already show jcenter might be faster than US's google mirror. 😇 
Let me do benchmark later.

PR #364, opened 2019-01-10
Fix session list loading error caused because Dispatcher loses action

### LINKED ISSUE #157: Fix session list is not displayed
## Overview (Required)
- I don't know how to reproduce.

screenshot_20190105-184656

## Links
-

kafumi:
## Issue
- close #157

### Cause of the issue

- `Action.UserRegistered` dispatched from `UserActionCreator` was randomly lost in race condition
  - If `Action.UserRegistered` is lost, `UserStore.registered` doesn't become `true` and `SessionContentsActionCreator.refresh()` never be triggered from `SessionPagesFragment`
  - It make session pages keep empty
- The loss is caused because of the implementation of `Dispatcher.launchAndDispatch()`:

```kotlin
class Dispatcher @Inject constructor() {
    private val _actions = BroadcastChannel<Action>(Channel.CONFLATED)
    // ...

    fun launchAndDispatch(action: Action) {
        GlobalScope.launch {
            _actions.send(action)
        }
    }
```

- The actual class of `_actions` is [`ConflatedBroadcastChannel`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-conflated-broadcast-channel/)
- `launchAndDispatch()` uses `GlobalScope` to start coroutine. It means the coroutine runs in [default dispatcher](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/-default.html), which is a thread pool on JVM
- If `launchAndDispatch()` is called multiple times in a short time, `ConflatedBroadcastChannel.send()` can be simultaneously executed in multiple thread
  - In such case, the 2nd and following calls are ignored by `ConflatedBroadcastChannel` because the channel doesn't have buffer. See  https://github.com/Kotlin/kotlinx.coroutines/blob/1.0.1/common/kotlinx-coroutines-core-common/src/channels/ConflatedBroadcastChannel.kt#L227-L230
- `Action.UserRegistered` is sometimes dispatched just after `Action.SessionsLoaded` when launching app
  - It causes loss of `Action.UserRegistered`

### Solution

- Make sure that `ConflatedBroadcastChannel.send()` is called in main thread not to lose any action

## Links
N/A

## Screenshot
Before | After
:--: | :--:
svid_20190111_021847_1 | svid_20190111_021643_1

### DIFF
--- a/frontendcomponent/androidcomponent/build.gradle
+++ b/frontendcomponent/androidcomponent/build.gradle
@@ -8,6 +8,7 @@ apply from: rootProject.file('gradle/android.gradle')
 
 dependencies {
     api project(":model")
+    api project(":ext:android-extension")
     api Dep.AndroidX.design
     api Dep.AndroidX.coreKtx
     implementation Dep.Kotlin.stdlibJvm
--- a/frontendcomponent/androidcomponent/src/main/java/io/github/droidkaigi/confsched2019/dispatcher/Dispatcher.kt
+++ b/frontendcomponent/androidcomponent/src/main/java/io/github/droidkaigi/confsched2019/dispatcher/Dispatcher.kt
@@ -1,6 +1,7 @@
 package io.github.droidkaigi.confsched2019.dispatcher
 
 import io.github.droidkaigi.confsched2019.action.Action
+import io.github.droidkaigi.confsched2019.ext.android.CoroutinePlugin
 import kotlinx.coroutines.GlobalScope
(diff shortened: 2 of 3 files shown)


panpanini:
👀

takahirom:
Thank! I will add issue that add test for this.

PR #367, opened 2019-01-11
Fixed files to be stored

jmatsu:
## Issue
- None

## Overview (Required)
- The current master failed due to a test but the results were not stored successfully
- I specified a test file but CircleCI required a directory which contains test results

## Links
-

PR #371, opened 2019-01-11
Fixed test files to be stored

jmatsu:
## Issue
- None

## Overview (Required)
- The current master failed due to a test but the results were not stored successfully
- I specified a test file but CircleCI required a directory which contains test results

## Links
-


jmatsu:
See the comments on the test results from https://github.com/DroidKaigi/conference-app-2019/pull/367

PR #376, opened 2019-01-11
Detect locale changed and refresh session page

### LINKED ISSUE #187: Fix category tags are displayed in previous languages when switching languages from system settings
## Overview (Required)
- Actually, after starting the application in Japanese setting, changing it to English will be like this

### Session
session

### Filter
filter

gen0083:
## Issue
- close #187 

## Overview (Required)
- When user changes their locale, views are not updated

So I implemented dynamic broadcast receiver in SystemActionCreator, and subscribing SystemProperty and dispatch refresh action if it was updated.

But i'm concerned with this approach especially registering broadcast receiver in ActionCreator. (I'm not familiar with multi-module project and flux manner, sorry...)

I hope to hear your honest opinions, thanks.

## Links
- https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner

## Screenshot
Before | After
:--: | :--:
before | after


takahirom:
LGTM!  👍 
his is just my preference, How about creating an onstart onstop function in the App class and calling observeSystemProperty() and unobserveSystemProperty() from there?

takahirom:
Sorry, I write sample code. 🙇 
Can you understand what I say ..? 🙇 

```

open class App : DaggerApplication() {
...
        ProcessLifecycleOwner.get().lifecycle.addObserver(systemActionCreator)
...
    @OnLifecycleEvent(Lifecycle.Event.ON_CREATE)
    private fun onCreated() {
        systemActionCreator.observeSystemProperty()
    }

    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
    private fun onDestroyed() {
        systemActionCreator.unobserveSystemProperty()
    }
```

gen0083:
Oh... sorry for my miss-understanding.  🙇 
You mean OnLifecycleEvent moves to App class from SystemActionCreator.. right?

takahirom:
Exactly👍

gen0083:
💪

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
```diff
@@ -30,12 +33,17 @@ open class App : DaggerApplication() {
...
-        systemActionCreator.setupSystem()
+        setupSystemActionCreator()
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_CREATE)
+    private fun onProcessCreated() {
+        systemActionCreator.observeSystemProperty()
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
```
takahirom: Can you use ON_START ON_STOP?

> ON_DESTROY will never be dispatched

https://developer.android.com/reference/android/arch/lifecycle/ProcessLifecycleOwner

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
```diff
@@ -77,6 +88,15 @@ open class App : DaggerApplication() {
         EmojiCompat.init(config)
     }
 
+    private fun setupSystemActionCreator() {
+        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
```
takahirom: Can you move this to onCreate? Because it can be used by other action creator.

gen0083: OK, I fix it :)

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
gen0083: If we use ON_START/ON_STOP, we cannot catch ACTION_LOCALE_CHANGED when app is background...
However, if ON_DESTROY is used, the broadcast receiver will not be unregistered. 🤔 

I implemented other version using BroadcastReceiver in AndroidManifest.xml, but this pattern got time-lagged until receive ACTION_LOCALE_CHANGED...

takahirom: Please let me think.🤔

gen0083: I guess re-checking SystemProperty in SessionPagesFragment is better than using dynamic broadcast receiver for fixing issue #187.
Something like this:

```kotlin
class SessionPagesFragment : DaggerFragment() {

override fun onStart() {
  ...
 systemActionCreator.checkSystemPropertyUpdated()
}
```

I don't know how to check current store state from SystemActionCreator...

takahirom: Well.. onStart in App, we can SystemActionCreator.refresh(). Now, We are using a changed extension function.
That function use `distinct()` in livedata-ktx. So probably it is OK.

takahirom: I added sample code 🙏 

```
open class App : DaggerApplication() {
...
        ProcessLifecycleOwner.get().lifecycle.addObserver(this)
...
    @OnLifecycleEvent(Lifecycle.Event.ON_START)
    private fun onStart() {
        systemActionCreator.refresh()
    }
```

gen0083: Thanks for the sample code.  😄 
I am working on it now, but It cannot reload as well. Please give me a little more time... 🙏

gen0083:
some improve added (some conversation is outdated sorry... 🙇 ):

- use lifecycle ON_START/ON_STOP to register dynamic broadcast receiver.
- to detecting locale change while application is in background, check locale action when application come back to front.
- add new Action `SystemLanguageChanged` for reloading Views

some places depend to SystemProperty (especially Lang), for example:
- https://github.com/DroidKaigi/conference-app-2019/blob/3e5e6e9b471918ab915c7670e8cf17eb09e7b40a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt#L140
- https://github.com/DroidKaigi/conference-app-2019/blob/3e5e6e9b471918ab915c7670e8cf17eb09e7b40a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/item/SpeechSessionItem.kt#L86
- etc

It is hard to check systemStore and update text each point, so I decided re-load MainActivity.

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
gen0083: I changed to use ON_START/ON_STOP, so broadcast receiver was unregistered when application go to background.
To detect ACTION_LOCALE_CHANGE while app is background, check lang when application come back to front.

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/ui/MainActivity.kt
```diff
@@ -85,6 +87,16 @@ class MainActivity : DaggerAppCompatActivity() {
...
             }
         }
+        systemStore.systemLanguageChanged.changed(this) {
+            Timber.debug { "detect language changed to $it" }
+            forceReloadCurrentDestination()
+        }
+    }
+
+    private fun forceReloadCurrentDestination() {
+        if (!navController.popBackStack(R.id.main, true)) {
```
gen0083: I don't have confidence as to whether this is OK.
Please let me know if there is a better way 🙏

takahirom:
Well, I rethink about this.
I want to say sorry 🙇 
Probably we should use the configuration change of Android.
When changing the language, the configuration changes occur.  And Activity is recreated.
But SystemAction is slow because of posting handler.
So probably we should use `defaultLang()` instead of `systemStore.lang`.
What do you think?

gen0083:
No way. I am sorry I sent you complicated PR too. 🙇 

This PR contains these points:

1. listening ACTION_LOCALE_CHANGE
2. update SystemProperty when language changed
3. reload Activity when language changed


> Probably we should use the configuration change of Android.

This is about point3? (trigger to reload Activity?)

> So probably we should use defaultLang() instead of systemStore.lang.

and this is about `systemActionCreator.refreshLangIfLocaleChanged(systemStore.lang)`?

takahirom:
All 😭 
I think we do no need to have the `lang` in systemProperty.(Currently, we don't need SystemAction). 
Because we don't use it. 🤔

gen0083:
Ah.. I felt this way was not good... 😇 

You said "use defaultLang() instead of systemStore.lang." means this?

https://github.com/DroidKaigi/conference-app-2019/blob/3e5e6e9b471918ab915c7670e8cf17eb09e7b40a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/item/SpeechSessionItem.kt#L86

`this@with.message.text = message.getByLang(systemStore.lang) ` -> `this@with.message.text = message.getByLang(defaultLang()) `

If so, I agree it is more simply way than this PR.
Well, I should close this PR and send new one is better?

takahirom:
> You said "use defaultLang() instead of systemStore.lang." means this?

Exactly

> Well, I should close this PR and send new one is better?

Either way is fine, I will leave it to you.

gen0083:
OK, I close this and send new one.
Thank you @takahirom  reviewing for a long time  🙇

PR #380, opened 2019-01-11
Updated ConstraintLayout

NUmeroAndDev:
## Issue
- None

## Overview (Required)
- Updated ConstraintLayout. 

I think, if someone may use MotionLayout it may be updated to [2.0.0-alpha03](https://androidstudio.googleblog.com/2018/12/constraintlayout-200-alpha-3.html). 🤔 

## Links
- https://androidstudio.googleblog.com/2018/08/constraintlayout-113.html


takahirom:
> I think, if someone may use MotionLayout it may be updated to 2.0.0-alpha03. 🤔

If we need that, I want to introduce it 👍

takahirom:
LGTM 🆙

PR #381, opened 2019-01-11
Introduce the coroutine context for action dispatch and fixed the race condition happened in only the test environment

jmatsu:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)

- Testcases using Dispatcher and ReadChannel#toLiveData are unstable due to threading.
- Introduced the coroutine context only for sending events and the current thread based coroutine context.
- This would be able to improve the performance because sending events do not depend on the main thread.

You can reproduce the test failure by the code below.

```
val breaker: (LoadingState?) -> Unit = { Thread.sleep(3000) }
val observer = mockk<(LoadingState?) -> Unit>(relaxed = true)

sessionsStore.loadingState.changedForever(breaker)
sessionsStore.loadingState.changedForever(observer)
```

*What happened?*

These tests blocked the test thread and could handle coroutine stuff correctly. However, they asserted stuff on another *thread* only in **test code**.
In this architecture, `ReadChannel#toLiveData` uses the main coroutine context to process events. And also, it's replaced with Default coroutine context in the testing environment. Default coroutine context is not a single-orientation context so sending actions and processing observers of LiveData will be executed concurrently. As a result, Observer's processing couldn't suspend the test thread. 

## Note
- This also removes placeholder unit tests

## Links
-


jmatsu:
@takahirom 
I'd like to hear your opinion on the naming of CoroutinePlugin's members. They are from Dispatchers but their types are cast-ed to CoroutineContext. Therefore, FooDispatcher is not suitable. And also, they and Action's Dispatcher are confusing. 🤔

frontendcomponent/androidtestcomponent/build.gradle
```diff
@@ -10,7 +10,6 @@ dependencies {
     implementation Dep.Kotlin.stdlibJvm
 
     api Dep.Dagger.androidSupport
-    implementation Dep.Kotlin.stdlibJvm
```
jmatsu: duplicated 🤣

feature/announcement/src/test/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStoreTest.kt
```diff
@@ -35,7 +35,7 @@ class AnnouncementStoreTest {
         dispatcher.dispatch(Action.AnnouncementLoadingStateChanged(LoadingState.LOADING))
         dispatcher.dispatch(Action.AnnouncementLoadingStateChanged(LoadingState.LOADED))
 
-        verifySequence {
+        verifyOrder {
```
jmatsu: This verification should check the order but not sequence.

feature/announcement/src/test/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStoreTest.kt
```diff
@@ -22,7 +22,7 @@ class AnnouncementStoreTest {
 
     @Before fun setUp() {
         MockKAnnotations.init(this, relaxUnitFun = true)
-        CoroutinePlugin.mainDispatcherHandler = { Dispatchers.Default }
+        CoroutinePlugin.mainDispatcherHandler = { TestActionDispatcher() }
```
jmatsu: mainDispatcherHandler will be used to process events of LiveData. So this means processing them will be done on the current thread. i.e. the test thread

jmatsu:
ah sorry, let me reconsider the context for action dispatcher...

jmatsu:
Let me wait for merging https://github.com/DroidKaigi/conference-app-2019/pull/411 

After we've talked in person, it should check the source and/or destination of channel streaming so #411 's verification is correct.

jmatsu:
The race condition has been fixed by #411

PR #395, opened 2019-01-12
[WIP] Create Xcode project

### LINKED ISSUE #389: Prepare Xcode project file using Kotlin Multi Platform
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- In order to achieve #132 , we need Xcode project to build iOS application.
- Maybe appropriate place of project files is in `frontend/ios` .
- This Xcode project uses Kotlin Multiplatform modules under `:data` , and will write other sources codes in Swift (Maybe it's hard to implement each screen like `:feature` modules in Kotlin Multiplatform modules...)
    - All ViewControllers are in single Xcode project.
    - We can use Cocoapods and Carthage as package manager.

kikuchy:
## Issue
- close #389 

## Overview (Required)
- Xcode project to build iOS app contains `:model` and `:data:api_impl` .
    - Open "frontend/ios/DroidKaigi 2019.xcodeproj" with Xcode and you can build app! 🎉 
- (not yet) Make some screen to be base writing codes.

## Problem

- Kotlin's suspended functions are not available in Framework header file.
    - So for example, `DroidKaigiApi#getSessions` can't call from Swift/Obj-C.
    - We should make converter suspended function to callback-style function or returning Promise like object function.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


model/src/commonMain/kotlin/SessionType.kt
```diff
@@ -7,7 +7,7 @@ enum class SessionType(val id: String, val isFavoritable: Boolean) {
     Codelabs("codelabs", true),
     FiresideChat("fireside_chat", false),
     Lunch("lunch", false),
-    Break("break", false),
+    BreakTime("break", false),
```
kikuchy: Kotlin object named `Break` converted into "break" at Framework's header files.  
Header files are C's headers, so it causes compile error because `break` is keyword of C.

takahirom:
😍 
image

data/api/build.gradle
```diff
@@ -18,7 +18,7 @@ kotlin {
     sourceSets {
         commonMain.dependencies {
             api Dep.Kotlin.stdlibCommon
-            api project(":model")
+            implementation project(":model")
```
kikuchy: `api` configuration propagates declarations to followed projects.  
It causes compile error in `:frontend:ios-combined` because `srcDirs` and dependencies have same class declarations.

It have no problem to change `api` to `implementation`.

kikuchy:
Oh, i found unnecessary gradle tasks. I'll remove it.

kikuchy:
Done!

PR #411, opened 2019-01-12
Refactor Store tests

takahirom:
## Overview (Required)
- The store's test is not in order, just checking if you can take the value and get it.
- This issue is needed for creating test issues

### DIFF
--- a/feature/announcement/src/main/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStore.kt
+++ b/feature/announcement/src/main/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStore.kt
@@ -16,7 +16,7 @@ class AnnouncementStore @Inject constructor(
     val loadingState: LiveData<LoadingState> = dispatcher
         .subscribe<Action.AnnouncementLoadingStateChanged>()
         .map { it.loadingState }
-        .toLiveData(LoadingState.LOADING)
+        .toLiveData(LoadingState.INITIALIZED)
     val announcements: LiveData<List<Announcement>> = dispatcher
         .subscribe<Action.AnnouncementLoaded>()
         .map { it.announcements }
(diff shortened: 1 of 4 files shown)


feature/announcement/src/main/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStore.kt
```diff
@@ -16,7 +16,7 @@ class AnnouncementStore @Inject constructor(
     val loadingState: LiveData<LoadingState> = dispatcher
         .subscribe<Action.AnnouncementLoadingStateChanged>()
         .map { it.loadingState }
-        .toLiveData(LoadingState.LOADING)
```
jmatsu: 👍

feature/announcement/src/test/java/io/github/droidkaigi/confsched2019/announcement/ui/store/AnnouncementStoreTest.kt
```diff
@@ -31,30 +31,26 @@ class AnnouncementStoreTest {
         val observer = mockk<(LoadingState?) -> Unit>(relaxed = true)
 
         announcementStore.loadingState.changedForever(observer)
+        coVerify { observer(LoadingState.INITIALIZED) }
```
jmatsu: I'm curious so let me ask one question, is this `coVerify` needed? Didn't `verify` work then?

takahirom: I will check that👍

takahirom: I confirm that. It is not needed. So I changed to verify! Thanks 👍

PR #412, opened 2019-01-12
Implemented fonts on about screen

NUmeroAndDev:
## Issue
- None

## Overview (Required)
- Implemented fonts on session detail screen.  
- Added "lekton" into dictionaries/dic.xml .  
- Apply code formatter.(It is set to format automatically when saving files.)

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
👀 
Thank you for applying font 🙇

feature/about/src/main/java/io/github/droidkaigi/confsched2019/about/ui/item/AboutItem.kt
```diff
@@ -16,12 +17,23 @@ class AboutItem(
     override fun getLayout(): Int = R.layout.item_about
 
     override fun bind(binding: ItemAboutBinding, position: Int) {
+        // clickListener is null when this item is version item
+        val isVersionItem = clickListener == null
```
takahirom: When adding item, if that person do not know this spec, it will cause bug.
Please add isLaketonCheckText field to AboutItem 🙏

NUmeroAndDev: @takahirom 
Thanks reply.
Please wait as it fixes. 🙇

PR #422, opened 2019-01-12
Enable org.gradle.parallel for faster build

miga9:
## Issue
There's no issues filed for this.

## Overview (Required)
I enabled `org.gradle.parallel` to take advantage of the multi-module setup. The build will run tasks in parallel where possible.

The amount of improved build time depends on the spec of your machine, but it should be faster in general.

Gradle scan reports
Enabled: https://scans.gradle.com/s/y5azqvabdv5yq
Disabled: https://scans.gradle.com/s/rz3r5yqbwt7ji

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -10,7 +10,7 @@ org.gradle.jvmargs=-Xmx4096m -XX:MaxPermSize=2048m -XX:+HeapDumpOnOutOfMemoryErr
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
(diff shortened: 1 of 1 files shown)


miga9:
I think there must be something flaky on this test that failed above. 
`io.github.droidkaigi.confsched2019.announcement.ui.store.AnnouncementStoreTest.loadingState`

I just force-pushed to trigger another check. (curious to know whether there is a re-test/re-run command that is set up in this project's CircleCi integration.)

takahirom:
I don't know that! Thanks 👍 

> take advantage of the multi-module setup. The build will run tasks in parallel where possible.

takahirom:
Gradle scan reports are very cool 🆒

miga9:
@takahirom Thanks for looking at it so fast :)
I updated the Overview to include the steps I took for the benchmark.

PR #424, opened 2019-01-13
Fixed of design sponsor screen

### LINKED ISSUE #420: Sponsor list screen is different from Gallery
## Kind (Required)

- Other

 ## Overview (Required)

The following items seem to be different.
- Card corner.
- Not divider.

Before | Current app
:--: | :--:
 |

NUmeroAndDev:
## Issue
- close #420 

## Overview (Required)
- Fixed of sponsor screen design.
  - Card corner changed to 16dp.
  - Added divider.
  - Fixed title text style.
  - Did code formatter.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/session/src/main/res/drawable/shape_dashed_line.xml
+++ b/feature/session/src/main/res/drawable/shape_dashed_line.xml
@@ -7,6 +7,6 @@
         android:color="@color/gray3"
         android:width="1dp"
         android:dashGap="3dp"
-        android:dashWidth="1dp"
+        android:dashWidth="3dp"
         />
 </shape>
--- a/feature/sponsor/src/main/java/io/github/droidkaigi/confsched2019/sponsor/ui/SponsorFragment.kt
+++ b/feature/sponsor/src/main/java/io/github/droidkaigi/confsched2019/sponsor/ui/SponsorFragment.kt
@@ -101,7 +101,7 @@ class SponsorFragment : DaggerFragment() {
                 sponsor.toItem(category)
             }
         )
-        if (isLastItem.not()) {
+        if (!isLastItem) {
             setFooter(DividerItem())
         }
         setHideWhenEmpty(true)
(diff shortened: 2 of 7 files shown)


feature/session/src/main/res/drawable/shape_dashed_line.xml
```diff
@@ -7,6 +7,6 @@
         android:color="@color/gray3"
         android:width="1dp"
         android:dashGap="3dp"
-        android:dashWidth="1dp"
+        android:dashWidth="3dp"
```
takahirom: Thanks 👍

feature/sponsor/src/main/java/io/github/droidkaigi/confsched2019/sponsor/ui/SponsorFragment.kt
```diff
@@ -88,18 +89,21 @@ class SponsorFragment : DaggerFragment() {
...
 
-    private fun SponsorCategory.toSection() = Section().apply {
+    private fun SponsorCategory.toSection(isLastItem: Boolean) = Section().apply {
         setHeader(HeaderItem(category.title))
         addAll(
             sponsors.map { sponsor ->
                 sponsor.toItem(category)
             }
         )
+        if (isLastItem.not()) {
```
takahirom: We do not use `.not()` in our current policy. 🙏 I think that it is okay to set up an issue if you are concerned.

NUmeroAndDev: @takahirom 
Fixed not to use `.not`.  https://github.com/DroidKaigi/conference-app-2019/pull/424/commits/08b795ef4daa9874d020db6034d4a0411bd274b9
Please review again 🙇 

Can you tell me why policy is not using `.not`?

takahirom: There is no particularly big reason. I think that this is pretty much due to preference, such as hard to see if denial later comes. Right now we are unifying with a policy not to use .not.

NUmeroAndDev: @takahirom 
I understood.
Thank you for answering. 😇

PR #426, opened 2019-01-13
Fix issue #187 use defaultLang instead of systemStore.lang

### LINKED ISSUE #187: Fix category tags are displayed in previous languages when switching languages from system settings

gen0083:
## Issue
- close #187 

## Overview (Required)
- when user changes device's language some `Chip` s are not updated until reloading/re-binding

This PR changes:

- use `defaultLang()` instead of `systemStore.lang`

## Links
- #376 

## Screenshot
Before | After
:--: | :--:
before | after


takahirom:
Thank you for tackling to this issue!

PR #441, opened 2019-01-13
Improve SessionsItemDecoration performance

### LINKED ISSUE #439: Scroll session page got a lot of memory allocation (probably performance issue)
**Overview (Required)**
- In SessionPageFragment, It seems that useless memory allocation has occurred when scrolling through the session list
- Memory usage is increasing even if it does not change in appearance
- There seems to be room for improvement in the SessionItemDecoration class

**Steps To Reproduce**

1. run application
1. open profiler view in Android Studio and start profiling session
1. scroll session list to downside
1. reach the bottom of session list
1. try to scroll down more and more then you can see memory allocation increasing

**Expected behavior**
avoid unnecessary memory allocation if we can

**Screenshots**
performance

**Env:**
 - Device: PH-1
 - OS: 9
 - branch master

gen0083:
## Issue
- close #439 

## Overview (Required)
- calculating session time on each onDrawOver that causes unnecessary memory allocation. (it is higher cost needed to calculate DateTime in onDrawOver)

## Links
-

## Screenshot
Before | After
:--: | :--:
performance | after

before: increase memory usage about 500MB and more
after: memory usage about 260MB

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
@@ -10,6 +10,7 @@ import android.util.SparseArray
 import android.view.View
 import androidx.core.content.ContextCompat
 import androidx.core.content.res.ResourcesCompat
+import androidx.core.util.contains
 import androidx.core.util.forEach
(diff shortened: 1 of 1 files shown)


gen0083:
> Is it work when change sessions by the filter?

Oh... it is completely outside of my brain... 😇 
I'll fix it 💪

takahirom:
Thank you for finding and fixing a critical leak!

PR #442, opened 2019-01-13
Disable Crashlytics if the apk has been built on the local environment

### LINKED ISSUE #433: Disable Crashlytics if the apk has been built on the local environment
## Kind (Required)

- New feature

 ## Overview (Required)

- The current Crashlytics setting is enabled in every environment including CI, your PC, and my PC. However, we should collect only crashes from apks which have been built on CI.
- `isCI` flag is available in build.gradle, and it will be true only on CI environment.

 ## Links	

-

akihito-okada:
## Issue
- close #433

## Overview (Required)
- The current Crashlytics setting is enabled in every environment including CI, your PC, and my PC. However, we should collect only crashes from apks which have been built on CI.
- isCI flag is available in build.gradle, and it will be true only on CI environment.


frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
```diff
@@ -81,6 +81,9 @@ open class App : DaggerApplication() {
     }
 
     private fun setupCrashlytics() {
+        if (!BuildConfig.IS_BUILT_ON_CI) {
+            return
+        }
         Fabric.with(this, Crashlytics())
```
jmatsu: The approach looks correct but Crashlytics needs to be initialized anyway, and it can disable itself from the code or AndroidManifest metadata. 
Could you please use `Crashlytics.Builder#disabled` or *firebase_crashlytics_collection_enabled metadata*?

frontend/android/build.gradle
```diff
@@ -29,6 +29,7 @@ android {
         resConfigs "ja"
         // For Android Studio Gradle sync
         applicationIdSuffix ".debug"
+        buildConfigField "boolean", "IS_BUILT_ON_CI", String.valueOf(isCi)
```
jmatsu: 👍

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
akihito-okada: Thank you for your review. 
I got it. I will try it !

akihito-okada:
@jmatsu 
I changed to initialize Crashlytics by firebase_analytics_collection_enabled.
Please review my change again.

akihito-okada:
@jmatsu @takahirom 
Still have problems? If there are any problems yet, please give me feedback

frontend/android/src/main/AndroidManifest.xml
```diff
@@ -14,7 +14,7 @@
         >
         <meta-data
             android:name="firebase_analytics_collection_enabled"
```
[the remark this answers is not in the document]
akihito-okada: @jmatsu I'm sorry. I misunderstood about parameter names. 
And thank you for your advice. 
I modified source code again and checked whether or not Crashlytics is disabled from LogCat.
Please confirm it.

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/App.kt
```diff
@@ -28,7 +26,6 @@ open class App : DaggerApplication() {
     override fun onCreate() {
         super.onCreate()
 
-        setupCrashlytics()
```
jmatsu: I didn't know CrashlyticsInitProvider could initialize itself automatically.

PR #449, opened 2019-01-14
iOS: Fetch sessions as model

takahirom:
## Overview (Required)
- Now, we can use session data.
- We can open and confirm it by opening frontend/ios with Xcode.

## Screenshot
image

### DIFF
--- a/buildSrc/src/main/java/dependencies/Dep.kt
+++ b/buildSrc/src/main/java/dependencies/Dep.kt
@@ -67,7 +67,7 @@ object Dep {
     }
 
     object Kotlin {
-        val version = "1.3.10"
+        val version = "1.3.11"
         val stdlibCommon = "org.jetbrains.kotlin:kotlin-stdlib-common:$version"
         val stdlibJvm = "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$version"
         val coroutinesVersion = "1.0.1"
--- a/buildSrc/src/main/java/dependencies/Dep.kt
+++ b/buildSrc/src/main/java/dependencies/Dep.kt
@@ -67,7 +67,7 @@ object Dep {
     }
 
     object Kotlin {
-        val version = "1.3.11"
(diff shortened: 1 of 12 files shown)


takahirom:
Let me merge for calling to iOS engineers

takahirom:
If you can comment here, we will give you a PR for improvement separately!

data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/MainLoopDispatcher.kt
```diff
@@ -0,0 +1,79 @@
...
+import kotlinx.coroutines.Runnable
+import platform.darwin.DISPATCH_TIME_NOW
+import platform.darwin.dispatch_after
+import platform.darwin.dispatch_async
+import platform.darwin.dispatch_get_main_queue
+import platform.darwin.dispatch_time
+import kotlin.coroutines.CoroutineContext
+
+@UseExperimental(InternalCoroutinesApi::class)
+object MainLoopDispatcher : CoroutineDispatcher(), Delay {
```
kikuchy: Awesome!!!!!!!! 🎉✨㊗️💪🔥

buildSrc/src/main/java/dependencies/Dep.kt
```diff
@@ -107,11 +107,13 @@ object Dep {
     }
 
     object Ktor {
-        val version = "1.0.1"
+        val version = "1.1.1"
         val clientCommon = "io.ktor:ktor-client-core:$version"
         val clientAndroid = "io.ktor:ktor-client-okhttp:$version"
+        val common = "io.ktor:ktor-client-json:$version"
```
kikuchy: `commonJson` is more better I think.

takahirom: Probably we should use `jsonCommon` 👍

PR #492, opened 2019-01-15
Fixed that about screen app version is not correct app version

### LINKED ISSUE #476: About screen app version is not correct app version
## Kind (Required)

- Other

 ## Overview (Required)

- It is not the app version. just strings.xml
- We should Add app version to versions.kts and use it in about/build.gradle 

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/about/src/main/res/values/strings.xml#L12

kirimin:
## Issue
- close #476

## Overview (Required)
- define versionName and versionCode to Versions.kts
- use it to build.gradle and about screen

## Links
-

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/buildSrc/src/main/java/dependencies/Versions.kt
+++ b/buildSrc/src/main/java/dependencies/Versions.kt
@@ -9,5 +9,5 @@ private object Versions {
     private val versionPatch = 0
     val androidVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
 
-    val androidVersionName = "1.0.0"
+    val androidVersionName = "$versionMajor.$versionMinor.$versionPatch"
 }
--- a/buildSrc/src/main/java/dependencies/Versions.kt
+++ b/buildSrc/src/main/java/dependencies/Versions.kt
@@ -7,7 +7,7 @@ private object Versions {
     private val versionMajor = 1
     private val versionMinor = 0
     private val versionPatch = 0
-    val androidvVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
+    val androidVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
 
-    val androidvVersionName = "1.0.0"
+    val androidVersionName = "1.0.0"
 }
--- a/buildSrc/src/main/java/dependencies/Versions.kt
+++ b/buildSrc/src/main/java/dependencies/Versions.kt
@@ -3,4 +3,11 @@ package dependencies
 private object Versions {
     val androidCompileSdkVersion = 28
     val androidMinSdkVersion = 21
+
+    private val versionMajor = 1
+    private val versionMinor = 0
+    private val versionPatch = 0
+    val androidvVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
+
+    val androidvVersionName = "1.0.0"
 }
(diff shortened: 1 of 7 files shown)


panpanini:
👀

feature/about/src/main/java/io/github/droidkaigi/confsched2019/about/ui/item/AboutItem.kt
```diff
@@ -10,7 +10,8 @@ import io.github.droidkaigi.confsched2019.about.databinding.ItemAboutBinding
 
 class AboutItem(
     @StringRes private val name: Int,
-    @StringRes private val description: Int,
+    @StringRes private val description: Int? = null,
```
panpanini: is there a reason to keep both `description` and `descriptionString` ?

buildSrc/src/main/java/dependencies/Versions.kt
```diff
@@ -3,4 +3,11 @@ package dependencies
 private object Versions {
     val androidCompileSdkVersion = 28
     val androidMinSdkVersion = 21
+
+    private val versionMajor = 1
+    private val versionMinor = 0
+    private val versionPatch = 0
+    val androidvVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
```
panpanini: it looks like maybe a typo 👀 should it be `androidVersionCode` instead?

buildSrc/src/main/java/dependencies/Versions.kt
```diff
@@ -3,4 +3,11 @@ package dependencies
...
+
+    val androidvVersionName = "1.0.0"
```
panpanini: also this should be `androidVersionName` 👍

feature/about/src/main/java/io/github/droidkaigi/confsched2019/about/ui/item/AboutSection.kt
```diff
@@ -46,10 +48,16 @@ class AboutSection @Inject constructor(
...
-                    description = R.string.about_version_name,
+                    descriptionString = activity.getVersionName(),
                     isLektonCheckText = true
                 )
             )
         )
     }
+
+    private fun Context.getVersionName() = try {
+        packageManager.getPackageInfo(packageName, 0).versionName
```
panpanini: instead of getting from `PackageManager`, what about [adding it to strings using `resValue`](https://www.tanelikorri.com/tutorial/android/set-variables-in-build-gradle/)?

buildSrc/src/main/java/dependencies/Versions.kt
kirimin: OMG, Sorry!
I fixed it!

feature/about/src/main/java/io/github/droidkaigi/confsched2019/about/ui/item/AboutSection.kt
kirimin: That great!
Thank you for advice!!

buildSrc/src/main/java/dependencies/Versions.kt
```diff
@@ -3,4 +3,11 @@ package dependencies
 private object Versions {
     val androidCompileSdkVersion = 28
     val androidMinSdkVersion = 21
+
+    private val versionMajor = 1
+    private val versionMinor = 0
+    private val versionPatch = 0
+    val androidVersionCode = versionMajor * 10000 + versionMinor * 100 + versionPatch
+
+    val androidVersionName = "1.0.0"
```
panpanini: ah, sorry I missed it before! what about using the versionMajor/minor/patch here as well?
``` kotlin
val androidVersionName = "$versionMajor.$versionMinor.$versionPatch"
```

kirimin: Awesome💕

kirimin:
Thank you for Reviewing!!

PR #501, opened 2019-01-16
Use carthage bootstrap instead of carthage update

takahirom:
## Overview (Required)
- Title says everything


takahirom:
let me merge

PR #504, opened 2019-01-16
Enable gzip compression when api calls

rnakano:
## Issue
- None

## Overview (Required)
I think using HTTP with gzip compression is a good way to reduce network traffic.

OkHttp adds `Content-Encoding: gzip` header automatically. However, to enable gzip in GoogleAppEngine, [it must also be included in User-Agent.](https://cloud.google.com/appengine/docs/standard/go/how-requests-are-handled#response_compression)

## Links
- https://cloud.google.com/appengine/docs/standard/go/how-requests-are-handled#response_compression

## Screenshot

Before ⬇️ 
 

After ⬇️ 

I confirmed that the traffic of the `timetable` decreased to 1/3 by the Android Profiler.

### DIFF
--- a/data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/ApiComponent.kt
+++ b/data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/ApiComponent.kt
@@ -14,7 +14,7 @@ internal fun generateHttpClient(): HttpClient {
     val version = NSBundle.mainBundle.objectForInfoDictionaryKey("CFBundleShortVersionString") as String
     return HttpClient(Ios) {
         install(UserAgent) {
-            agent = "official-app-2019/$version"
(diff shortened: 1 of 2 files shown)


panpanini:
> However, to enable gzip in GoogleAppEngine, it must also be included in User-Agent.

📝 thats really useful information!! thanks!

panpanini:
I wonder about iOS as well, is it possible to do something for this in `ApiComponent.kt` as well?
https://github.com/DroidKaigi/conference-app-2019/blob/master/data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/ApiComponent.kt#L17

rnakano:
Of course! I will also modify iOS code.

rnakano:
I confirmed that the traffic was reduced even on iOS 😃 

⬇️ Profiler Screenshot

| Before | After |
|:-------|:------|
|  |  |

PR #505, opened 2019-01-16
Fix flickering when back to timeline with back key

### LINKED ISSUE #235: Flicker when you back to timeline with back key
## Overview (Required)

Steps
1. Open app
2. Open Drawer
3. Tap any menu item (e.g. `DroidKaigiとは` )
4. Tap back key
5. Screen filled with red color once, and back to current screen (e.g. `DroidKaigiとは` ) immediately, after that, fade transition is started (e.g. fade transition from `DroidKaigiとは` to `タイムライン`). `Screen filled with red color once` is this issue.

## Links
-

gen0083:
## Issue
- close #235 

## Overview (Required)
- navigation with `setupWithNavController` handling navigation in `NavigationUI` class
- fade animation is used (nav_default_anim_enter.xml etc)
- flickering may occur when applying fade animation to both the enter side and the exit side
- this PR apply fade animation on only one side (exit only)

In `NavigationUI` class, set transition was hard-coded.
I cannot find way to customize it.
So I prepare animation resources and overwrite it.

Other solution is handling ourselves instead of `setupWithNavController`.

I could not decide which method to take, so I will send a PR with fewer modifications

image

## Links
- https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/navigation/ui/src/main/res/anim/nav_default_enter_anim.xml
- https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/navigation/ui/src/main/java/androidx/navigation/ui/NavigationUI.java#75

## Screenshot
Before | After
:--: | :--:
before | after


gen0083:
I'm sorry, I thought that I should use fade...
What is the specification of transition?

takahirom:
I'm sorry I was worrying a bit. The designer of this application seems to want to do this.
https://material.io/design/motion/understanding-motion.html#usage

takahirom:
Can you try` handling ourselves instead of setupWithNavController`? 🙏

ymnder:
Probably you can do it like this, customized NavigationUI :) 
```
class CustomizedNavigationUI {
    fun setupWithNavController(
            @NonNull navigationView: NavigationView,
            @NonNull navController: NavController
     ) { //...original code
     }

    fun onNavDestinationSelected(
        @NonNull item: MenuItem,
        @NonNull navController: NavController, popUp: Boolean
    ): Boolean {
        val builder = NavOptions.Builder()
            .setLaunchSingleTop(true)
            .setEnterAnim(R.anim.nav_enter_anim) // set the custom anim here
            .setExitAnim(R.anim.nav_exit_anim)
            .setPopEnterAnim(R.anim.nav_pop_enter_anim)
            .setPopExitAnim(R.anim.nav_pop_exit_anim)
        if (popUp) {
            builder.setPopUpTo(findStartDestination(navController.graph)!!.id, false)
        }
        val options = builder.build()
        try {
            //TODO provide proper API instead of using Exceptions as Control-Flow.
            navController.navigate(item.itemId, null, options)
            return true
        } catch (e: IllegalArgumentException) {
            return false
        }
    }
    //...
}
```

gen0083:
@ymnder thank you for your suggestion 😄 

@takahirom  I'll try it a bit more so I closed this once and send again later 💪

Issue #65, opened 2018-12-13
Add session survey screen


takahirom:
## Overview (Required)
- We need to create the session survey

## Links
- Design https://gallery.io/projects/MCHbtQVoQ2HCZd9Pbpex4PVP/files/MCHtA7U1iMGr64hSKBFaITmFolw2KJ84irU (Please send request)

PR #507, opened 2019-01-16
Add empty session survey screen

futabooo:
## Issue
- #65

## Overview (Required)
- Add empty session survey screen

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
If you want to use session title, session speaker... I think you can use Parcelize like this.
https://aakira.app/blog/2018/12/kotlin-mpp-android-parcelable/

PR #508, opened 2019-01-16
Add error callback for iOS in common module

### LINKED ISSUE #493: Add error callback for iOS in common module
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- We should add error callback for iOS
- We can use `try catch` for `suspend function` and call callback.

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/master/data/api-impl/src/commonMain/kotlin/io/github/droidkaigi/confsched2019/data/api/KtorDroidKaigiApi.kt#L39

amay077:
## Issue
- close #493

## Overview (Required)
- Add ``onError`` callback to ``DroidKaigiApi.getSessions``
- Receive onError in iOS(swift) and implement display alert

## Screenshot
Before | After
:--: | :--:
N/A | 

I'm not sure that how to raise exception in ``DroidKaigiApi.getSessions``.
I tried turn offline my iPhone but ``getSessions`` returns success(empty response).
So I change [ApiEndpoint](https://github.com/DroidKaigi/conference-app-2019/blob/a5dc5133cf1e1ad4eaad435c4816aa43a2e6cb8d/data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/ApiEndpoint.kt#L5) to invalid url then I can confirmed call error callback.

### DIFF
--- a/frontend/ios/DroidKaigi 2019/SessionsViewController.swift
+++ b/frontend/ios/DroidKaigi 2019/SessionsViewController.swift
@@ -27,6 +27,18 @@ final class SessionsViewController: UIViewController {
             }
             .disposed(by: disposeBag)
         
+        viewModel.error
+            .bind {[weak self] (errorMessage) in
+                if let errMsg = errorMessage {
+                    let alert = UIAlertController(title: "ERROR", message: errMsg, preferredStyle: UIAlertController.Style.alert)
+                    let closeButton = UIAlertAction(title: "Close", style: UIAlertAction.Style.cancel, handler: nil)
+                    alert.addAction(closeButton)
+                    
+                    self?.present(alert, animated: true, completion: nil)
+                }
+            }
+            .disposed(by: disposeBag)
+
         self.viewModel = viewModel
     }
 }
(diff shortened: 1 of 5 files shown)


frontend/ios/DroidKaigi 2019/SessionsViewController.swift
```diff
@@ -27,6 +27,18 @@ final class SessionsViewController: UIViewController {
             }
             .disposed(by: disposeBag)
         
+        viewModel.error
+            .bind {[weak self] (errorMessage) in
```
takahirom: Please do not ask me because I do not know iOS. Is it possible to be called more than once, such as when viewDidLoad returns from the background? Will there be errors twice at that time?

amay077: As far as I know, ``viewDidLoad`` is similar to ``onPostCreate`` in Android.

* [iOS と Android で画面表示時のコールバックを比較する - Qiita](https://qiita.com/amay077/items/52a0b0da97fe455abc08)
* [AndroidエンジニアのためのiOSのUIViewControllerのライフサイクルとAndroidのActivityのライフサイクル比較 - Qiita](https://qiita.com/kgmyshin/items/b54e84d71b3406cd6e79)

So ``viewDidLoad`` does not called when returns from background.

amay077:
Thanks, this is my first commit by swift :confetti_ball:

PR #510, opened 2019-01-17
Fix ktlint warning and append target paths

sakebook:
## Issue
- No issue

## Overview (Required)
- Append ktlint path tests and MPP folders
- Fix run ktlintFormat error


e10dokup:
👀

PR #511, opened 2019-01-18
Update iOS README

ikesyo:
Hi I'm @ikesyo, a maintainer of Carthage 😄 Thanks for using it!

## Overview (Required)

There should be no need to let Carthage build macOS variant and so on.


takahirom:
🆒

PR #519, opened 2019-01-18
#167 Add tabState to store and subscribe it.

### LINKED ISSUE #167: Sometime unneeded animation myplan tab
## Overview (Required)
- You can see gif
animation

maripiyoko:
## Issue
- close #167

## Overview (Required)
- store tabState
- watch tabState value change
- when tabState becomes IDLE (= means swipe action finished) set BottomSheet to expand

## Note
Actually I wanted to cancel the user touch event but couldn't since below actions occur almost same time.
- swiping horizontally for tab switch
- swiping vertically for BottomSheet to collapse

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
👀

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionPageFragment.kt
```diff
@@ -124,6 +125,11 @@ class SessionPageFragment : DaggerFragment() {
                 bottomSheetBehavior.state = BottomSheetBehavior.STATE_EXPANDED
             }
         }
+        sessionPagesStore.tabState.changed(viewLifecycleOwner) {
+            if (it == ViewPager.SCROLL_STATE_IDLE) {
+                bottomSheetBehavior.state = BottomSheetBehavior.STATE_EXPANDED
+            }
```
maripiyoko: at first I wanted to set `BottomSheetBehavior.isDraggable = false` for `ViewPager.SCROLL_STATE_DRAGGING` here.
but the time is too short and BottomSheet down action is already triggered.. so it's not worked 😢

takahirom:
Sorry, I don't understand this issue yet. It seems occurred 🤔 
unneedanimation

takahirom:
Probably I find a solution 👀  
Can you check this? 🙇 

In BottomSheetBehavior

image
image

takahirom:
Sorry, My code has a bug.
It is becoming difficult to close BottomSheet when fling.

PR #532, opened 2019-01-18
Fix memory leak associated with groupAdapter

tomoya0x00:
## Issue

none

## Overview (Required)
- Fix memory leak associated with groupAdapter
  - Avoid holding groupAdapter in Fragment's field
  - Make weak reference to groupAdapter in SessionsItemDecoration

## Links

none

## Screenshot

none

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetFavoriteSessionsFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetFavoriteSessionsFragment.kt
@@ -59,8 +59,6 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
         InjectedViewModelProviders.of(requireActivity()).get(sessionPagesStoreProvider)
     }
 
-    private val groupAdapter = GroupAdapter<ViewHolder<*>>()
-
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
@@ -74,6 +72,7 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
 
     override fun onActivityCreated(savedInstanceState: Bundle?) {
         super.onActivityCreated(savedInstanceState)
+        val groupAdapter = GroupAdapter<ViewHolder<*>>()
         binding.sessionsRecycler.apply {
             adapter = groupAdapter
             addItemDecoration(
@@ -96,7 +95,7 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
         binding.sessionsRecycler.addOnScrollListener(object : RecyclerView.OnScrollListener() {
             override fun onScrolled(recyclerView: RecyclerView, dx: Int, dy: Int) {
                 super.onScrolled(recyclerView, dx, dy)
-                applyTitleText()
+                applyTitleText(groupAdapter)
             }
         })
 
@@ -123,7 +122,7 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
                 }
 
             groupAdapter.update(items)
-            applyTitleText()
+            applyTitleText(groupAdapter)
             binding.shouldShowEmptyStateView = items.isEmpty()
             if (items.isEmpty()) {
                 binding.sessionsListHeaderShadow.isVisible = false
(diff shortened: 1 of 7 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
```diff
@@ -19,11 +19,13 @@ import io.github.droidkaigi.confsched2019.session.R
 import io.github.droidkaigi.confsched2019.session.ui.item.SessionItem
 import io.github.droidkaigi.confsched2019.timber.debug
 import timber.log.Timber
+import java.lang.ref.WeakReference
 
 class SessionsItemDecoration(
     val context: Context,
-    val groupAdapter: GroupAdapter<*>
+    groupAdapter: GroupAdapter<*>
```
jmatsu: It seems this class wants only item count of an adapter, and the direction is restricted to vertical.
So we don't need passing an adapter to this class, and then we can use `RecyclerView#getRowCountForAccessibility` or `RecyclerView#getAdapter()?.getItemCount()` instead.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetFavoriteSessionsFragment.kt
```diff
@@ -152,7 +151,7 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
         super.onDestroyView()
     }
 
-    private fun applyTitleText() {
+    private fun applyTitleText(groupAdapter: GroupAdapter<*>) {
```
jmatsu: We can use a computed property like below:

```
val groupAdapter get() = binding. sessionsRecycler.adapter as GroupAdapter<ViewHolder<*>>()
```

In addition, if this method doesn't require the concrete class, i.e. GroupAdapter, then `binding. sessionsRecycler.adapter?.itemCount ?: 0` is enough.

feature/sponsor/src/main/java/io/github/droidkaigi/confsched2019/sponsor/ui/SponsorFragment.kt
```diff
@@ -82,12 +82,17 @@ class SponsorFragment : DaggerFragment() {
...
+            setupSponsorsLayout(it, groupAdapter)
+        }
 
         sponsorActionCreator.load()
     }
 
-    private fun setupSponsorsLayout(sponsorCategories: List<SponsorCategory>) {
+    private fun setupSponsorsLayout(
+        sponsorCategories: List<SponsorCategory>,
+        groupAdapter: GroupAdapter<*>
```
jmatsu: ditto.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
jmatsu: BTW, `RecyclerView$State#getItemCount` cannot be used in this case because it looks up the prelayout state.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetFavoriteSessionsFragment.kt
```diff
@@ -59,7 +59,8 @@ class BottomSheetFavoriteSessionsFragment : DaggerFragment() {
         InjectedViewModelProviders.of(requireActivity()).get(sessionPagesStoreProvider)
     }
 
-    private val groupAdapter = GroupAdapter<ViewHolder<*>>()
+    private val groupAdapter
+        get() = binding.sessionsRecycler.adapter as GroupAdapter<*>
```
tomoya0x00: To suppresses the warning of `Unchecked cast`, use `GroupAdapter <*>` instead of `GroupAdapter<ViewHolder<*>>`.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
tomoya0x00: Unfortunately, this class uses `GroupAdapter#getItem` in the `getSessionTime` method 😇

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetFavoriteSessionsFragment.kt
tomoya0x00: Thank you for your good suggestion! 🆒 
Changed to use computed property. 2511776

Unfortunately, this method uses `GroupAdapter#getItem` 😇

tomoya0x00:
@jmatsu Thank you for the review.
I've addressed all suggestions. Please take a look when you have time 🙏

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/SessionsItemDecoration.kt
jmatsu: Oh sorry, I was wrong. 😞

PR #537, opened 2019-01-19
Fix layout for BottomSheet

### LINKED ISSUE #167: Sometime unneeded animation myplan tab

maripiyoko:
## Issue
- close #167 

## Overview (Required)
- Use android:layout_height="wrap_content" for RecyclerView
- https://github.com/DroidKaigi/conference-app-2019/issues/167#issuecomment-455617087


takahirom:
There is still a problem of animating when moving to MyPlan after launching the application.
I think we should change this 👀 
In BottomSheetBehavior

image

PR #589, opened 2019-01-22
Blink the session detail when it is favorite

### LINKED ISSUE #549: Fix blink when favorite in session detail
## Kind (Required)

- Other

 ## Overview (Required)

- The title says everything.
- Probably it is occurred by [here](https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt#L208)

iaia:
## Issue
- close #549 

## Overview (Required)
- Blink session detail when the session was favorite

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
So, Sorry 🙇 🙇 
I want to fix blink.
I will fix that issue 🙇

PR #593, opened 2019-01-23
fix ios default local

numa08:
`NSLocal.preferredLanguages` returns Language IDs
https://developer.apple.com/documentation/foundation/nslocale/1415614-preferredlanguages
https://qiita.com/m1takahashi/items/e28858198ed8848a2738


takahirom:
completely understood
image

PR #595, opened 2019-01-23
Fix blink SessionDetail is blinking when favorite button clicked

### LINKED ISSUE #549: Fix blink when favorite in session detail

gen0083:
## Issue
- close #549 

## Overview (Required)
- Blinking was triggered that checking lineCount after setText (see figure1).
- This PR changes checking lineCount before setText
- and fix SpeakerItem equals method

figure1
figure1: slow motion

but at some points, there is a room of improve:

- ellipsis position is not exact same before situation (I think it is difficult to detect exact position before `setText`)
- It seems be better to split to CustomView such a draw logic

## Links
-

## Screenshot
Before | After
:--: | :--:
before_ellipsis_position | after_ellipsis_position
before | after

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
@@ -170,7 +170,7 @@ class SessionDetailFragment : DaggerFragment() {
         binding.speechSession = session
         val lang = defaultLang()
         binding.lang = lang
-        setupSessionDescription()
+        setupSessionDescription(session.desc)
         binding.timeZoneOffset = DateTimeSpan(hours = 9) // FIXME Get from device setting
 
         binding.sessionTitle.text = session.title.getByLang(lang)
@@ -187,8 +187,6 @@ class SessionDetailFragment : DaggerFragment() {
         session.message?.let { message ->
             binding.sessionMessage.text = message.getByLang(defaultLang())
         }
-
-        binding.sessionDescription.text = session.desc
 
         val speakerItems = session
             .speakers
@@ -234,20 +232,22 @@ class SessionDetailFragment : DaggerFragment() {
         }
     }
 
-    private fun setupSessionDescription() {
+    private fun setupSessionDescription(fullText: String) {
         val textView = binding.sessionDescription
         textView.doOnPreDraw {
+            // check the number of lines if set full length text (this text is not displayed yet)
+            textView.text = fullText
+            // if lines are more than prescribed value then collapse
             if (textView.lineCount > 5 && showEllipsis) {
                 val end = textView.layout.getLineStart(5)
                 val ellipsis = getString(R.string.ellipsis_label)
                 val ellipsisColor = ContextCompat.getColor(requireContext(), R.color.colorSecondary)
                 val onClickListener = {
                     TransitionManager.beginDelayedTransition(binding.sessionLayout)
-                    val session = binding.speechSession?.desc
-                    binding.sessionDescription.text = session
+                    textView.text = fullText
                     showEllipsis = !showEllipsis
                 }
-                val detailText = textView.text.subSequence(0, end - ellipsis.length)
+                val detailText = fullText.subSequence(0, end - ellipsis.length)
                 val text = buildSpannedString {
                     clickableSpan(onClickListener, {
                         append(detailText)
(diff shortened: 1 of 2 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -277,6 +280,15 @@ class SessionDetailFragment : DaggerFragment() {
         }, builderAction)
     }
 
+    private fun computeLineCount(fullText: String): Int {
```
jmatsu: We should use `TextUtils#ellipsize` or `measureText` to detect this.

`TextUtils#ellipsize` can ellipsize the given text actually so it's really reliable.
`getTextBounds` rounds values but `measureText` can treat values as float, so measureText has higher accuracy.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -277,6 +280,15 @@ class SessionDetailFragment : DaggerFragment() {
...
+        if (binding.sessionDescription.width == 0) return 0
+        val rect = Rect()
+        val paint = Paint()
+        paint.textSize = binding.sessionDescription.textSize
+        paint.getTextBounds(fullText, 0, fullText.length, rect)
+        return rect.width() / binding.sessionDescription.width
```
jmatsu: We should add +1 line to return # of lines to be rendered

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
jmatsu: I prefer TextUtils#ellipsize because of https://github.com/DroidKaigi/conference-app-2019/pull/595/files#diff-a45d9aed66a757a483bc9cdf2b196c9aR251

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/item/SpeakerItem.kt
```diff
@@ -71,7 +71,7 @@ class SpeakerItem @AssistedInject constructor(
         other as SpeakerItem
 
         if (speaker != other.speaker) return false
-        if (query == null || other.query == null || query != other.query) return false
+        if (query != other.query) return false
```
jmatsu: It seems `query == null && other.query == null` should be false in the previous code but this doesn't. Is it okay?

gen0083: Previous code was wrong. Because GroupAdapter updated SpeakerItem even if it have same contents. (SpeakerItem looked blinking with the previous code)

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
gen0083: Wow, I don't know such methods, thank you! 👍 
I'll improve it 💪

gen0083:
@jmatsu I apologize for misunderstanding.　🙇 
compute line count was not needed...

but using `TextUtils.ellipsize` is a good idea and it resolves another issue of displaying ellipsis text.
(It can be confirmed by opening the English session details screen)

another_issue

I'll fix it by another PR using `TextUtils.ellipsize` after this PR merged.
(or do it in this PR?)

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -234,20 +232,22 @@ class SessionDetailFragment : DaggerFragment() {
         }
     }
 
-    private fun setupSessionDescription() {
+    private fun setupSessionDescription(fullText: String) {
         val textView = binding.sessionDescription
         textView.doOnPreDraw {
+            // check the number of lines if set full length text (this text is not displayed yet)
+            textView.text = fullText
```
jmatsu: 👀

jmatsu:
> I'll fix it by another PR using TextUtils.ellipsize after this PR merged.

Sure, please! 🆒

PR #607, opened 2019-01-24
Fixed sponsor page could not be opened due to json strict mode

jmatsu:
## Issue
- None

## Overview (Required)

- The server response contains `kind` field so json mode should be non-strict.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
Gracias 🇪🇸

PR #611, opened 2019-01-24
Fix ellipsis position was shifted in some situation

### LINKED ISSUE #608: Ellipsize session descriptions by TextUtil instead of substring
## Kind (Required)

- Other

 ## Overview (Required)

- This issue was derived from https://github.com/DroidKaigi/conference-app-2019/pull/595
- @gen0083 thank you so much.

 ## Links	

-

gen0083:
## Issue
- close #608 

## Overview (Required)
- in some situation, ellipsis position is shifted

## discussion point

If 5th line (that is prescribed value decided to ellipsize) contains line feed, it is shifted even if using `TextUtils.ellipsize`
so I decided to remove line feed(`\n`) from 5th line and after.

Pros:
- ellipsis positioned at the end of 5th line

Cons:
- in some cases, break 5th line (see 4th before/after screenshot)

how do you think?

## Screenshot
Before | After
:--: | :--:
before_expected | after_expected
before_english | after_english
before_line_break | after_line_break
before_line_break_2 | after_line_break_2

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
@@ -3,6 +3,7 @@ package io.github.droidkaigi.confsched2019.session.ui
 import android.os.Bundle
 import android.text.SpannableStringBuilder
 import android.text.TextPaint
+import android.text.TextUtils
 import android.text.method.LinkMovementMethod
 import android.text.style.ClickableSpan
 import android.view.LayoutInflater
@@ -45,6 +46,8 @@ import io.github.droidkaigi.confsched2019.system.actioncreator.ActivityActionCre
 import io.github.droidkaigi.confsched2019.user.store.UserStore
 import io.github.droidkaigi.confsched2019.util.ProgressTimeLatch
 import javax.inject.Inject
+
+private const val ELIPSIS_LINE_COUNT = 5
 
 class SessionDetailFragment : DaggerFragment() {
     private lateinit var binding: FragmentSessionDetailBinding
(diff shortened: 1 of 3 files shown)


frontendcomponent/androidcomponent/src/main/res/values/strings.xml
```diff
@@ -15,7 +15,7 @@
     <string name="map_label">Floormap</string>
     <!-- Survey -->
     <string name="survey_label">Entire survey</string>
-    <string name="ellipsis_label">…続きを読む</string>
+    <string name="ellipsis_label">more</string>
```
takahirom: Nice catch 👍

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -242,16 +245,26 @@ class SessionDetailFragment : DaggerFragment() {
             // check the number of lines if set full length text (this text is not displayed yet)
             textView.text = fullText
```
jmatsu: Does this TextView have the exact width already? If so, we can simplify this processing like this.

```
  if (!showEllipsis) {
    textView.text = fullText
    return@doPreDraw
  }

  val noLineReturn = fullText.replace("\n".toRegex(), "")
  val ellipsized = TextUtils.ellipsize(
                    noLineReturn,
                    paint,
                    textView.width * ELIPSIS_LINE_COUNT,
                    TextUtils.TruncateAt.END
                )
  if (ellipsized == noLineReturn) { // no need to ellipsize
    textView.text = fullText
    return@onPreDraw
  }

  textView.text = // buildSpannedString based on elipsized
  ...
```

gen0083: Thank you reviewing. 😄 
I'll try to improve it (you mean use "early return", right?)

複数行どうするんだろうと思ってたら単に行数かければよかったんですね・・・。
もしかしたら禁則処理あたりでうまくいかないかもしれませんが、それも含めて試してみます。

jmatsu: For now, the current code looks good! so let me merge 👍 please send another PR if the change looks better. 🙏

PR #636, opened 2019-01-26
Fix max width of tags for iOS

woxtu:
## Issue
- None

## Overview (Required)
- In small devices (SE, 5s, etc.), long tags is sticking out from a table view

## Screenshot
Before | After
:--: | :--:
 |


frontend/ios/DroidKaigi 2019/Scenes/Session/TagsCollectionViewCell.swift
```diff
@@ -93,7 +93,6 @@ class TagsCollectionViewCell: UICollectionViewCell, Reusable {
         addSubview(label)
         label.snp.makeConstraints {
             $0.edges.equalToSuperview()
-            $0.width.lessThanOrEqualTo(230)
```
jmatsu: Doesn't it work if we make the right constraint of the label fit to the right of the cell?

jmatsu: Gotcha. Thank you so much!

woxtu:
🙌

PR #643, opened 2019-01-26
[ios]add ripple effect to tableViewCell

shcahill:
## Issue
- none

## Overview (Required)
- add ripple effect to Session list(TableViewCell) 
- using MaterialComponents for iOS

## Links
- https://github.com/material-components/material-components-ios/tree/develop/components/Ink

## Screenshot
Before | After
:--: | :--:
before | ripple


takahirom:
Looks great 👍

PR #645, opened 2019-01-27
Enable session list to session survey

takahirom:
## Overview (Required)
- You can go to the session survey from the session list.

## Screenshot

licecap


takahirom:
Let me merge  🙏

PR #648, opened 2019-01-27
Fix session detail survey button ripple and hide when service session

takahirom:
## Overview (Required)
- We should show ripple effect when touch survey button
- Currently, we can use session survey for service session. So we should remove it when service session.

## Screenshot
image
image


takahirom:
Let me merge

PR #654, opened 2019-01-27
Should not go to session detail when sessionType.supportDetail is false

takahirom:
## Overview (Required)
- Should not go to session detail when sessionType.supportDetail is false
- Currently, we can go session detail of lunch


takahirom:
Let me merge

PR #655, opened 2019-01-27
[Proposal] PagerTabStrip for iOS

koogawa:
I'm not very familiar with MVVM, therefore I might destroy your architecture. If so, please reject this PR 🙏 

## Overview (Required)
-  PagerTabStrip for iOS

demo

## Links
- https://github.com/xmartlabs/XLPagerTabStrip

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Thanks 👍 We should add this for AppStore review 🙏 🍎

PR #657, opened 2019-01-28
Store the generated universal apk as well

jmatsu:
## Issue
- Degrade since #650 

## Overview (Required)
- Need to store a universal apk as well

## Links
-


jmatsu:
Let me merge

PR #667, opened 2019-01-28
iOS tweaks for release

takahirom:
## Overview (Required)
- ios-combined -> ioscombined(can not use `-`)
- set VALID_ARCHS
- Change bundle name


takahirom:
Let me merge

PR #682, opened 2019-01-29
[iOS] Open link when speaker tapped in session detail view

417-72KI:
## Issue
- No Issue

## Overview (Required)
- When speaker tapped on session detail view and speaker has links for Twitter / GitHub / Blogs, 
App should open those links.

## Screenshot
ezgif com-resize


takahirom:
Sorry, probably that field will become always empty 🙇 
https://droidkaigi2019-dev.appspot.com/api/timetable

takahirom:
I think your feature is very useful. If we can use 😭

PR #687, opened 2019-01-30
Restore scroll position for TimetableLayoutManager

### LINKED ISSUE #616: save first visible item position or scroll position and restore it for TimeTableLayoutManager
## Kind (Required)

- Other

 ## Overview (Required)

- Title says everything

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/f49c824fd85934f656505904de4369036d8c51d7/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimeTableLayoutManager.kt#L10

## Assign
@MoyuruAizawa

MoyuruAizawa:
## Issue
- close #616 

## Overview (Required)
- Restore scroll position
- Rename  'TimeTableLayoutManager' -> 'TimetableLayoutManager'

## Links
none

## Screenshot
no change


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -52,6 +53,34 @@ class TimeTableLayoutManager(
         LEFT, TOP, RIGHT, BOTTOM
     }
 
+    private data class SaveState(val position: Int, val left: Int, val top: Int) : Parcelable {
```
jmatsu: We are using `kotlinx.parcelize` 👀 Could you please use it instead?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -64,35 +93,57 @@ class TimeTableLayoutManager(
...
-        var offsetX = parentLeft
-        for (columnNum in 0 until columns.size()) {
-            if (columns.indexOfKey(columnNum) == 0) anchor.leftColumn = columnNum
-            offsetX += fillColumnHorizontally(columnNum, 0, offsetX, parentTop, true, recycler)
-            if (offsetX > parentRight) {
-                anchor.rightColumn = columnNum
-                break
-            }
-        }
+        val firstVisibleView = if (childCount > 0) findFirstVisibleView() else null
```
jmatsu: Doesn't `findFirstVisibleView` return null when this has no children?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -64,35 +93,57 @@ class TimeTableLayoutManager(
...
+        val offsetX = saveState?.left ?: firstVisibleView?.let(this::getDecoratedLeft)
+        val offsetY = saveState?.top ?: firstVisibleView?.let(this::getDecoratedTop)
+        val period = saveState?.position?.let(periods::getOrNull)
```
jmatsu: Can we simplify like this?

`(saveState?.position ?: firstVisibleView?.adapterPosition)?.let(periods::getOrNull)`

I'm afraid `periods::getOrNull` can return null as well. Which of mine or yours is correct?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -64,35 +93,57 @@ class TimeTableLayoutManager(
...
+            ?: firstVisibleView?.adapterPosition?.let(periods::getOrNull)
+        detachAndScrapAttachedViews(recycler)
+        if (offsetX != null && offsetY != null && period != null)
+            relayoutChildren(offsetX, offsetY, period, recycler)
```
jmatsu: Don't we need to round offsets like `Math.max(offsetX, parentLeft)` and/or `Math.max(offsetY, parentTop)`?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -227,6 +283,48 @@ class TimeTableLayoutManager(
...
+            offsetX += fillColumnHorizontally(
+                columnNumber,
+                startPositionInColumn.positionInColumn,
+                offsetX,
+                offsetY,
+                true,
+                recycler
+            )
+
+            if (i == columnCount - 1) i = 0 else i++
```
jmatsu: I'm not sure whether or not there is a huge device though, the following sounds better. What do you think ?

```
...
var iterateCount = 0
while (... && !shouldLoopHorizontally && iterateCount < columnCount) {
  ...
  iterateCount++
}
```

MoyuruAizawa: after thinking about it, it will not work on large screen devices.
maybe, columns will be drawn like this...
`|3|4|5|6|7|8|0|1|2|3|4|`

I will rewrite.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
MoyuruAizawa: in this case, the first child view has negative `left`(and `top`)

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -227,6 +258,49 @@ class TimeTableLayoutManager(
...
+    }
+
+    private fun relayoutChildren(startX: Int, startY: Int, topPeriod: Period, recycler: Recycler) {
+        var offsetX = startX
+        anchor.leftColumn = topPeriod.columnNumber
+        val columnCount = columns.size()
+        val indexOfFirstColumn = columns.indexOfKey(topPeriod.columnNumber)
+        (indexOfFirstColumn until columnCount)
+            .plus(if (shouldLoopHorizontally && indexOfFirstColumn > 0) (0 until indexOfFirstColumn) else emptyList())
+            .forEach {
```
MoyuruAizawa: how about this?

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/TabularFormSessionPageFragment.kt
```diff
@@ -63,34 +63,34 @@ class TabularFormSessionPageFragment : DaggerFragment() {
 
         val groupAdapter = GroupAdapter<ViewHolder<*>>()
         binding.tabularFormSessionsRecycler.apply {
-            addItemDecoration(TimeTableTimeLabelDecoration(context, groupAdapter))
-            addItemDecoration(TimeTableRoomLabelDecoration(context, groupAdapter))
-            layoutManager = TimeTableLayoutManager(
+            addItemDecoration(TimetableTimeLabelDecoration(context, groupAdapter))
```
MoyuruAizawa: rename decorations too.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimeTableDividerDecoration.kt
```diff
@@ -1,107 +0,0 @@
...
-import androidx.core.view.children
-import androidx.recyclerview.widget.RecyclerView
-import com.xwray.groupie.GroupAdapter
-import io.github.droidkaigi.confsched2019.session.R
-import io.github.droidkaigi.confsched2019.session.ui.item.TabularServiceSessionItem
-import io.github.droidkaigi.confsched2019.session.ui.item.TabularSpacerItem
-import io.github.droidkaigi.confsched2019.session.ui.item.TabularSpeechSessionItem
-import java.util.concurrent.TimeUnit
-
-class TimeTableDividerDecoration(
```
MoyuruAizawa: unneeded.

feature/session/build.gradle
```diff
@@ -3,10 +3,16 @@ import dependencies.Dep
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-kapt'
+apply plugin: 'kotlin-android-extensions'
 apply plugin: 'androidx.navigation.safeargs'
 
 apply from: rootProject.file('gradle/android.gradle')
-android.dataBinding.enabled = true
+android {
```
jmatsu: ah sorry, they were not enabled in this module. Thanks. 👍

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
jmatsu: The logic looks good. 💯

I guess `anchor.rightColumn` should be assigned if this loop finishes without through https://github.com/DroidKaigi/conference-app-2019/pull/687/files#diff-535b121f65cd5ca3a47e0df6b54a7ab4R301.
So it's better to define this range as a local variable once, maybe.

```
val range = (indexOfFirstColumn until columnCount)
            .plus(
                if (shouldLoopHorizontally && indexOfFirstColumn > 0) (0 until indexOfFirstColumn)
                else emptyList()
            )

range.forEach { ... }

anchor.rightColumn = requireNotNull(range.last())
```

MoyuruAizawa: `Anchor.rightColumn` is not always the last of the range.
If `columnWidth` is 100px, the range is `[2, 3, 4, 5, 0, 1]` and the width of RecyclerView is 500px,  `Anchor.rightColumn` is `0`

jmatsu: Yes, I know. Maybe you are saying about a case of https://github.com/DroidKaigi/conference-app-2019/pull/687/files#diff-535b121f65cd5ca3a47e0df6b54a7ab4R302
I meant a case when this loop finishes *without* through the branch.

jmatsu: @MoyuruAizawa

MoyuruAizawa: Ah, OK. _COMPLETELY UNDERSTOOD_
In case width of columns is less than width of RecyclerView, we cant't scroll. So, `Anchor.leftColumn` and `Anchor.rightColumn` is not referenced.
But I agree your suggestion. I fix it.

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLayoutManager.kt
```diff
@@ -227,6 +258,53 @@ class TimeTableLayoutManager(
...
+        var offsetX = startX
+        anchor.leftColumn = topPeriod.columnNumber
+        val columnCount = columns.size()
+        val indexOfFirstColumn = columns.indexOfKey(topPeriod.columnNumber)
+        val range = (indexOfFirstColumn until columnCount)
+            .plus(
+                if (shouldLoopHorizontally && indexOfFirstColumn > 0) (0 until indexOfFirstColumn)
+                else emptyList()
+            )
+        anchor.rightColumn = columns.keyAt(requireNotNull(range.last()))
```
jmatsu: Smart!

PR #689, opened 2019-01-30
[iOS] Change API endpoint according to build configuration

### LINKED ISSUE #674: Use the production API endpoint for iOS app as well
## Kind (Required)

- New feature

 ## Overview (Required)

- Change api endpoints based on Schema for iOS app
- https://github.com/DroidKaigi/conference-app-2019/blob/master/data/api-impl/build.gradle#L73 should be used for the release build

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/master/data/api-impl/src/iosMain/kotlin/io/github/droidkaigi/confsched2019/data/api/ApiEndpoint.kt#L3

hironytic:
## Issue
- close #674 

## Overview (Required)
- In addition to `src/iOSMain/kotlin`, one of
  `src/iOSMain/kotlinDebug`
   or
  `src/iOSMain/kotlinRelease`
  is also source directory now, according to Xcode configuration (Debug or Release)
- Move`ApiEndpoint.kt` to `kotlinDebug` and `kotlinRelease` , making it return appropriate endpoint

I couldn't find out the way to tell the configuration value to Kotlin source code. So I decided to switch  the whole source file itself.
`project.findProperty("XCODE_CONFIGURATION")` is used for detecting Xcode configuration. I don't know much about it 😇 but it is already used in `packForXCode` task in ioscombined/build.gradle.

Though I've also changed the api-impl/build.gradle, I think is not used for iOS build 🤔

### DIFF
--- a/data/api-impl/build.gradle
+++ b/data/api-impl/build.gradle
@@ -50,6 +50,13 @@ kotlin {
             }
         }
         iOSMain {
+            final String mode = project.findProperty("XCODE_CONFIGURATION")?.toUpperCase() ?: 'DEBUG'
+            if (mode == "RELEASE") {
+                kotlin.srcDirs += "src/iosMain/kotlinRelease"
+            } else {
+                kotlin.srcDirs += "src/iosMain/kotlinDebug"
+            }
+
             dependsOn commonMain
             dependencies {
                 implementation Dep.Ktor.clientIos
--- a/data/api-impl/build.gradle
+++ b/data/api-impl/build.gradle
@@ -50,13 +50,6 @@ kotlin {
             }
         }
         iOSMain {
-            final String mode = project.findProperty("XCODE_CONFIGURATION")?.toUpperCase() ?: 'DEBUG'
-            if (mode == "RELEASE") {
-                kotlin.srcDirs += "src/iosMain/kotlinRelease"
-            } else {
-                kotlin.srcDirs += "src/iosMain/kotlinDebug"
-            }
-
             dependsOn commonMain
             dependencies {
                 implementation Dep.Ktor.clientIos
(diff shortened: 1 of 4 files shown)


data/api-impl/src/iosMain/kotlinDebug/io/github/droidkaigi/confsched2019/data/api/ApiEndpoint.kt
```diff
@@ -1,5 +1,3 @@
 package io.github.droidkaigi.confsched2019.data.api
 
-// TODO: Replace with code getting endpoint from iOS application build configurations.
```
jmatsu: 👍

frontend/ioscombined/build.gradle
```diff
@@ -53,8 +53,14 @@ kotlin {
             }
         }
         iOSMain {
+            final String mode = project.findProperty("XCODE_CONFIGURATION")?.toUpperCase() ?: 'DEBUG'
             projectsList.forEach {
                 kotlin.srcDirs += "${project(it).projectDir}/src/iosMain/kotlin"
+                if (mode == "RELEASE") {
```
jmatsu: Sounds smart. 🆒

data/api-impl/build.gradle
```diff
@@ -50,6 +50,13 @@ kotlin {
             }
         }
         iOSMain {
+            final String mode = project.findProperty("XCODE_CONFIGURATION")?.toUpperCase() ?: 'DEBUG'
```
jmatsu: `projectsList` in ioscombined can handle this, so I guess this diff is not required. Didn't it work without this chunk? 👀

hironytic: Yes. It works without this.

hironytic:
> As you notied, could you please try removing the hunk of changes?

OK. Wait a minute.

PR #692, opened 2019-01-30
[iOS] Add floor map

### LINKED ISSUE #686: Floormap for iOS
## Kind (Required)

- Other

 ## Overview (Required)

- You can use these images. Please use that!
- https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/floormap/src/main/res/drawable-xxhdpi/ic_floor1.png
- https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/floormap/src/main/res/drawable-xxhdpi/ic_floor2.png

 ## Links	

-

417-72KI:
## Issue
- close #686

## Overview (Required)
- Add Floor map Screen
- Add link for floor map screen

Since I didn't enough time for creating menu, I add icon for floor map link on `rightBarButtonItems` temporary.

## Screenshot
animation


frontend/ios/DroidKaigi 2019/Scenes/FloorMap/FloorMapContentViewController.swift
```diff
@@ -0,0 +1,40 @@
...
+//
+//  Created by 417.72KI on 2019/01/31.
+//
+
+import UIKit
+import XLPagerTabStrip
+
+class FloorMapContentViewController: UIViewController {
+
+    var floor: Floor?
```
takahirom: I'm not good at Swift. But I think If always floor injected by the system. And field don't become null. When Kotlin code, We uses `Floor!` type (Kotlin `lateinit` property)

417-72KI: I'm afraid of crash by forgetting to set floor in caller view.
[FYI] Some projects set lint to give warning with `Hoge!` type (In swift, we call them **IUO**).

But in this case, you're right because it is *Logic Error* and app should crash.
So I will fix it.

417-72KI:
@takahirom 
Thanks for review!
 I answered and fixed it!

ShoMasegi:
Sorry for interfering.🙇‍♂️
I think we don't need `R.swift` for using only storyboard instantiate.
If you want to instantiate storyboard(ViewController), please look [StoryboardInstantiable+Extension](https://github.com/DroidKaigi/conference-app-2019/blob/master/frontend/ios/DroidKaigi%202019/Protocols/StoryboardInstantiable.swift).

I want to make this app lightweight.🙇‍♂️

frontend/ios/DroidKaigi 2019/Extensions/NSLayoutConstraint+Extension.swift
```diff
@@ -0,0 +1,14 @@
...
+//  NSLayoutConstraint+Extension.swift
+//  DroidKaigi 2019
+//
+//  Created by 417.72KI on 2019/01/31.
+//
+
+import UIKit
+
+extension NSLayoutConstraint {
+    static func activate(_ constraints: NSLayoutConstraint...) {
```
ShoMasegi: Forget to delete this `Extension`?

frontend/ios/DroidKaigi 2019/Scenes/MainViewController.swift
```diff
@@ -5,16 +5,23 @@
 //  Created by koogawa on 2019/01/28.
 //
 
+import RxSwift
 import UIKit
 import XLPagerTabStrip
 
 class MainViewController: ButtonBarPagerTabStripViewController, StoryboardInstantiable {
 
+    private let bag = DisposeBag()
```
ShoMasegi: this property is not necessary?

frontend/ios/DroidKaigi 2019/data/Floor.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Floor.swift
+//  DroidKaigi 2019
+//
+//  Created by 417.72KI on 2019/01/31.
+//
+
+import Foundation
+import class UIKit.UIImage
+
+struct Floor {
```
ShoMasegi: There is only `Floor1`, `Floor2`.
so I think it is better to use `enum`.

frontend/ios/DroidKaigi 2019/Scenes/MainViewController.swift
417-72KI: I forgot to remove that.

417-72KI:
@ShoMasegi Thanks for review!
I was afraid of typos, and also considered using resource files like *Localizable.string*.

frontend/ios/DroidKaigi 2019/Scenes/MainViewController.swift
417-72KI: I fixed it!

frontend/ios/DroidKaigi 2019/data/Floor.swift
ShoMasegi: In any case,I think we need to fix codes when floors changed.🤔

417-72KI: Surely... 🤔
I fixed it!

ShoMasegi:
@417-72KI 
Sorry for interfering.🙇‍♂️
It is nice idea, but we don't have so many strings.
so 'I' don't think to need it in this case.

417-72KI:
@ShoMasegi 
Okay, thanks for discussion!

takahirom:
@ShoMasegi Thank you for review!

PR #695, opened 2019-01-30
Open current session when first loading

### LINKED ISSUE #684: Show current time session when loading session first in Fragment
## Kind (Required)

- Other

 ## Overview (Required)

- We want to show the current time session when conference starts
- We want to scroll to the current session when first loading.
- As there are times when the timing is severe, I think it's good to have flags like this in the Fragment flag for this, like isInitializedScrollPosition.

Now we can use `scrollToCurrentSession ` 
https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt#L162

nacatl:
## Issue
- close #684

## Overview (Required)
- When sessions are loaded, if it is first time for the Fragment, scroll to current session.
- It will never scroll automatically at opened unless you kill the MainActivity

## Links
- none

## Screenshot
The time of the device was set 2019/02/08 15:00

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
@@ -132,6 +132,11 @@ class BottomSheetDaySessionsFragment : DaggerFragment() {
                 ?.session
                 ?.startDayText ?: return@changed
             binding.sessionsBottomSheetTitle.text = titleText
+
+            if (sessionPagesStore.sessionScrollAdjusted.value == false) {
+                scrollToCurrentSession()
+                sessionPagesActionCreator.dispatchSessionScrollAdjusted()
+            }
         }
         sessionPagesStore.filters.changed(viewLifecycleOwner) {
             binding.isFiltered = it.isFiltered()
@@ -153,13 +158,6 @@ class BottomSheetDaySessionsFragment : DaggerFragment() {
             if (SessionPage.pageOfDay(args.day) == it) {
                 scrollToCurrentSession()
             }
-        }
-
-        sessionPagesStore.firstLoadSessions.changed(viewLifecycleOwner) {
-            sessionPagesActionCreator.firstLoad()
-        }
-        sessionPagesStore.scrollWhenFirstLoaded.changed(viewLifecycleOwner) {
-            if (it) scrollToCurrentSession()
         }
     }
 
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
@@ -156,11 +156,10 @@ class BottomSheetDaySessionsFragment : DaggerFragment() {
         }
 
         sessionPagesStore.firstLoadSessions.changed(viewLifecycleOwner) {
-            if (sessionPagesStore.scrolledWhenFirstLoad.value == true) {
-                return@changed
-            }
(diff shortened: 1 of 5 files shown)


feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
```diff
@@ -152,6 +152,13 @@ class BottomSheetDaySessionsFragment : DaggerFragment() {
                 scrollToCurrentSession()
             }
         }
+
+        sessionPagesStore.firstLoadSessions.changed(viewLifecycleOwner) {
+            if (!sessionPagesStore.isFirstLoadScrolled) {
+                scrollToCurrentSession()
+                sessionPagesStore.isFirstLoadScrolled = true
```
takahirom: Can you do it by action? It will break Flux flow 😭

nacatl: That is right...I'm sorry... :bowing_man: 
I change it

nacatl:
@takahirom 
I changed to Action, but it is not worked...

Did I get using ActionCreator( or the other) wrong…?

nacatl:
Sorry,  force-push had fixed a lint and a little mistake...

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/BottomSheetDaySessionsFragment.kt
```diff
@@ -152,6 +154,13 @@ class BottomSheetDaySessionsFragment : DaggerFragment() {
                 scrollToCurrentSession()
             }
         }
+
+        sessionPagesStore.firstLoadSessions.changed(viewLifecycleOwner) {
+            sessionPagesActionCreator.firstLoad()
+        }
+        sessionPagesStore.scrollWhenFirstLoaded.changed(viewLifecycleOwner) {
+            if(it) scrollToCurrentSession()
```
nacatl:
```suggestion
            if (it) scrollToCurrentSession()
```

nacatl:
Separate [Loaded action ignition] and [receive action and scroll].
Probably fixed...

takahirom:
Sorry, can you check this?🙇
https://twitter.com/new_runnable/status/1090899653338443776?s=19

takahirom:
How about this?
https://github.com/DroidKaigi/conference-app-2019/commit/a192c281dd6beacdc9bcbfbcaf5d1dcadd3dd8e2

takahirom:
I think we can fix that by following codes.
But probably I will it after DroidKaigi

```
open class Store : ViewModel() {
    private val onClearHooks: MutableList<() -> Unit> = mutableListOf()
    fun addHook(hook: () -> Unit) {
        onClearHooks.add(hook)
    }

    override fun onCleared() {
        super.onCleared()
        onClearHooks.forEach { it() }
    }
}
```

```
@MainThread fun <T> ReceiveChannel<T>.toLiveData(
    store: Store,
    defaultValue: T? = null
): LiveData<T> {
    return object : LiveData<T>(), CoroutineScope by GlobalScope {
        init {
            if (defaultValue != null) {
                value = defaultValue
            }
            val job = launch(CoroutinePlugin.mainDispatcher) {
                for (element in this@toLiveData) {
                    postValue(element)
                }
            }
            store.addHook { job.cancel() }
        }
    }
}
```

nacatl:
I think you are right. It is good codes but will maybe make an effect a wide range...

By the way, how shall we merge a192c28?
May I cherrypick it?

takahirom:
Can you cherrypick? 🙏

nacatl:
Thank you for your support!!
:bowing_man:

PR #697, opened 2019-01-31
Update Requirements

takahirom:
## Overview (Required)
- Update Android Studio Kotlin Plugin version
- Add https://github.com/DroidKaigi/conference-app-2019/issues/690


takahirom:
🙏

PR #702, opened 2019-01-31
[Proposal] new feature to insert a calendar events

nukka123:
## Issue
- no issue.

## Overview (Required)

How about a new feature?
This PR proposes a feature to insert a calendar events on the session detail screen.

## Screenshot
Add Calendar Button (SessionDetail)  | Insert new events (Google Calendar) | Saved (Google Calendar)
:--: | :--: | :--:
 |  |

### DIFF
--- a/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
+++ b/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
@@ -111,6 +111,17 @@ class SessionDetailFragment : DaggerFragment() {
                         }
                     )
                 }
+                R.id.session_calendar -> {
+                    val session = binding.session ?: return@setOnMenuItemClickListener false
+                    if (session is SpeechSession) {
+                        activityActionCreator.openCalendar(
+                            session.title.getByLang(defaultLang()),
+                            session.room.name,
+                            session.startTime.unixMillisLong,
+                            session.endTime.unixMillisLong
+                        )
+                    }
+                }
             }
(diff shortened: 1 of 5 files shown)


e10dokup:
@nukka123 Thanks for your cool proposals! In this app, `SessionDetailFragment` handles not only `SpeechSession` but also `ServiceSession`. So when displaying `ServiceSession` (eg codelabs programs), nothing happens even if the icon is pressed on `OnMenuItemClickListener` as it is now. 

In my thought, there are two solutions.

- Hide `R.id.session_calendar` icon on displaying `ServiceSession`
- Enable to insert a calendar events even in `ServiceSession`

It would be great if you would do either. What are your thoughts on that?

nukka123:
@e10dokup 

>but also ServiceSession.
Oh! I see.

> Enable to insert a calendar events even in ServiceSession
Yes. It makes sense!

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/SessionDetailFragment.kt
```diff
@@ -113,13 +113,23 @@ class SessionDetailFragment : DaggerFragment() {
...
+                    when (session) {
+                        is SpeechSession -> {
+                            activityActionCreator.openCalendar(
+                                session.title.getByLang(defaultLang()),
+                                session.room.name,
+                                session.startTime.unixMillisLong,
+                                session.endTime.unixMillisLong
+                            )
+                        }
+                        is ServiceSession -> {
```
e10dokup: 👍

PR #726, opened 2019-02-01
Add TimetableLunchItemDecoration

### LINKED ISSUE #620: Overwrite lunch session for session time table by ItemDecoration
## Kind (Required)

- Proposal / Discussion

 ## Overview (Required)

- Currently, lunch is only HallA, But we want to spread it. It is hard to realize this with LayoutManager, so overwrite it with ItemDecoration.

image

 ## Links	

- https://github.com/DroidKaigi/conference-app-2019/blob/master/feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/TabularFormSessionPageFragment.kt#L61

itome:
## Issue
- close #620 

## Overview (Required)
- Implement TimetableLunchItemDecoration to overwrite luch row.
- Find lunch item from adapter and draw on the view in `onDrawOver`.

## Links
-

## Screenshot
Before | After
:--: | :--:
ezgif com-video-to-gif | ezgif com-video-to-gif 1


takahirom:
@MoyuruAizawa If you have any opinion for this PR. Please comment it 🙏

feature/session/src/main/java/io/github/droidkaigi/confsched2019/session/ui/widget/TimetableLunchDecoration.kt
```diff
@@ -0,0 +1,127 @@
...
+import com.xwray.groupie.GroupAdapter
+import com.xwray.groupie.Item
+import io.github.droidkaigi.confsched2019.model.SessionType
+import io.github.droidkaigi.confsched2019.session.R
+import io.github.droidkaigi.confsched2019.session.ui.item.TabularServiceSessionItem
+import io.github.droidkaigi.confsched2019.session.ui.item.TabularSpacerItem
+import java.util.concurrent.TimeUnit
+
+class TimetableLunchDecoration(
+    private val timeLabelWidth: Float,
```
takahirom: Why do you use parameter for it? ⛱ 
I think you can just use field like this.

```
class TimetableLunchDecoration(context: Context, groupAdapter: GroupAdapter<*>) {
    private val timeLabelWidth: Float = context.resources.getDimension(R.dimen.tabular_form_time_label_width)
```

takahirom:
Can you fix this? 👀 
lunch

itome:
I fixed two bugs
- Text appears out of lunch item
- Lunch decoration does'nt appear in ROOM5, ROOM6, in Day1

And I did little hacky for fixing it.
I added lunch items for all rooms before adding groupie Items.

takahirom:
Can you fix the conflict? 👀

takahirom:
Thanks 👍

PR #737, opened 2019-02-02
Separated classes which depend only on debug and staging build types correctly

jmatsu:
## Issue
- none

## Overview (Required)
- Stetho was missing on release build
- As for staging, some of the integrations didn't work because of missing code

## How to test

- [x] Confirm the following gradle commands work

- ./gradlew clean assembleDebug
- ./gradlew clean assembleStaging
- ./gradlew clean assembleRelease


jmatsu:
Let me merge/

PR #750, opened 2019-02-02
[Proposal] Add the WiFi configuration by one click

jmatsu:
## Issue
- None

## Overview (Required)

- This is a proposal of a new function to add a WiFi configuration from App easily.
- I introduced a preference for it to do this for now.

## Discussion points

- How to abstract WifiManager

I've created new modules which are device and device-impl to hide the platform specific implementation from MPP aspect. However, I'm also thinking it's enough to make the Android-device specific implementation in `repository-impl` and make it MPP.

Conclusion: Don't use repository modules but device modules are okay. System module will use device module directly.

- Which screen is preferable

I've chosen the preference screen but I guess the about screen sounds also good.

Conclusion: Use About screen

## Links
-

## Screenshot
1 | 2
:--: | :--: 
 |


data/repository-impl/src/main/java/io/github/droidkaigi/confsched2019/data/repository/DataWifiConfigurationRepository.kt
```diff
@@ -0,0 +1,19 @@
+package io.github.droidkaigi.confsched2019.data.repository
+
+import io.github.droidkaigi.confsched2019.data.device.WifiManager
+import io.github.droidkaigi.confsched2019.model.WifiConfiguration
+import kotlinx.coroutines.coroutineScope
+import javax.inject.Inject
+
+class DataWifiConfigurationRepository @Inject constructor(
```
takahirom: I think that you do not have to use the Repository except when you want to cache it. 👀

feature/settings/src/main/java/io/github/droidkaigi/confsched2019/settings/ui/SettingsFragment.kt
```diff
@@ -1,11 +1,82 @@
...
+
+    @Inject lateinit var systemStore: SystemStore
+    @Inject lateinit var systemActionCreator: SystemActionCreator
+    @Inject lateinit var activityActionCreator: ActivityActionCreator
 
-class SettingsFragment : PreferenceFragmentCompat() {
     override fun onCreatePreferences(savedInstanceState: Bundle?, rootKey: String?) {
         setPreferencesFromResource(R.xml.preferences, rootKey)
+
+        requireNotNull(findPreference(getString(R.string.pref_key_add_wifi)))
```
takahirom: 📝

feature/settings/src/main/java/io/github/droidkaigi/confsched2019/settings/ui/SettingsFragment.kt
```diff
@@ -1,11 +1,82 @@
...
+            .setOnPreferenceClickListener {
+                true.also {
```
takahirom: Sorry, Why is this needed? 🤔

feature/system/src/main/java/io/github/droidkaigi/confsched2019/system/actioncreator/ActivityActionCreator.kt
```diff
@@ -60,6 +62,16 @@ class ActivityActionCreator @Inject constructor(val activity: FragmentActivity)
         activity.startActivity(intent)
     }
 
+    fun copyText(label: String, text: String): Boolean {
+        val cpManager =
+            ContextCompat.getSystemService(activity, ClipboardManager::class.java) ?: return false
```
takahirom: I think you can use `ShowProcessingMessage` instead of returning Boolean. What do you think? 👀 

`dispatcher.launchAndDispatch(Action.ShowProcessingMessage(Message.of("...")))`

jmatsu: Messages we want to show depend on the situation, so let me create new action for copy text.

frontend/android/src/main/java/io/github/droidkaigi/confsched2019/ui/MainActivity.kt
```diff
@@ -93,8 +95,18 @@ class MainActivity : DaggerAppCompatActivity() {
...
+                return@changed
+            }
+
             val messageStr: String = when (message) {
-                is Message.ResourceIdMessage -> getString(message.messageId)
+                is Message.ResourceIdMessage -> {
+                    if (message.stringArgs.isEmpty()) {
+                        getString(message.messageId)
+                    } else {
+                        getString(message.messageId, *message.stringArgs)
```
takahirom: 👏

PR #765, opened 2019-02-03
Fix when back to tabular session, the scroll progress is reset

### LINKED ISSUE #741: When back to tabular session, the scroll progress is reset
## Kind (Required)

- Other

 ## Overview (Required)

- Open App
- Goto tabular session
- Goto session detail
- Back to the tabular session. Scroll state is reset 😭

sekiya9311:
## Issue
- close #741 

## Overview (Required)
- The place to create TabularFormSessionPageFragment is only onCreate()
- Reuse TabularFormSessionPageFragment of displayed before goto session detail.

## Links
- None

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
If we do it in onCreate(). Does it make it more simple? 
If you try it, It is OK 🙇 
https://twitter.com/new_runnable/status/1082960999689273345

takahirom:
Thanks, it looks good! 
This knowledge seems to be very important when making with a single activity. 👍

PR #787, opened 2019-02-04
Add versionCode offset to avoid accidental versionCode duplication 

jmatsu:
## Issue
- None

## Overview (Required)
- PlayStore cannot accept the same versionCode once we upload so I've added versionCode offset which is kinda *reserved* version space.


jmatsu:
Let me merge.

PR #788, opened 2019-02-04
[Proposal] Use new announcement icons

jmatsu:
## Issue
- none

## Overview (Required)
- We don't have vector files for the current icons, so it might be better to use material icons instead to prepare for iOS icons.

## Links
- https://drive.google.com/open?id=1R6lEMomFePkfWtmoCwfIfR2c7N3qc9h7

## Screenshot
Before | After
:--: | :--:
 |


jmatsu:
Got an approval internally so let me merge this.

PR #804, opened 2019-02-05
Use Travis to publish an apk to alpha track

jmatsu:
## Issue
- close 

## Overview (Required)
- Unfortunately https://github.com/Triple-T/gradle-play-publisher/ cannot be used because our assembling does not depend on assemble tasks 😢 
- So we need to create a publisher logic by ourselves.

## Links
-


jmatsu:
Let me merge

PR #811, opened 2019-02-06
Change app name for non-release builds.

matsujun:
## Issue
- no issue

## Overview (Required)
- Current app has same name for debug/staging builds and release build, but their package name is not same. In launcher app screen, we cannot identify which is debug/staging/release build.

## Links
-

## Screenshot
Build | Before | After
:--: |:--: | :--:
Debug |  | 
Staging |  |


takahirom:
LGTM 🐞

PR #825, opened 2019-02-11
Use screen lifecycle based on #637

takahirom:
## Issue
- close #578 

## Overview (Required)
- Use ViewModel scope for action creator
- By #637, we can create ViewModel scope.


corecomponent/androidcomponent/src/main/java/io/github/droidkaigi/confsched2019/widget/PageViewModel.kt
```diff
@@ -0,0 +1,17 @@
+package io.github.droidkaigi.confsched2019.widget
+
+import androidx.lifecycle.ViewModel
+
+class PageViewModel : ViewModel() {
```
jmatsu: Isn't it enough if ViewModel extends LifecycleOwner directly? 🤔

